PR #1, opened 2019-09-04
Avoid globally changing thread behaviour

janko:
There might be code that's relying on the default behaviour of threads,
that an exception in threads won't propagate to the main thread. So it's
safer that Ferrum changes this behaviour only for threads it creates.


route:
Thanks, looks good to me! I'll merge after fixing some failing specs on old ruby and capybara.

PR #3, opened 2019-09-05
Support MacOS, and some more regional formats

TGRGIT:
Ctrl+A does not work on MacOS, and the output of the date format is changes in different regions.

### DIFF
--- a/spec/browser/input_spec.rb
+++ b/spec/browser/input_spec.rb
@@ -175,7 +175,8 @@ module Ferrum
 
       it "sets contenteditable's content" do
         input = browser.at_css("#filled_div")
-        input.focus.type([:Ctrl, "A"], :Backspace, "new text")
+        input.evaluate('this.innerHTML = "";')
+        input.focus.type(:Backspace, "new text")
         expect(input.text).to eq("new text")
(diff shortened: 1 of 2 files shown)


route:
Thanks for fixing it on mac, I would prefer in general to do it w/o JS, so maybe better to use something like `Ferrum.mac? ? %i[Alt Left] : %i[Ctrl Left]`

TGRGIT:
Updated to remove evaluate. I can confirm the tests pass on macos, but I don't have a linux image handy to test with.

spec/browser/input_spec.rb
```diff
@@ -5,6 +5,7 @@
 module Ferrum
   describe "Browser::API::Input" do
     let!(:browser) { Browser.new(base_url: @server.base_url) }
+    let!(:select_all) { Ferrum.mac? ? %i[Alt shift Right] : %i[Ctrl A] }
```
route: That shouldn't be with exclamation mark, because we don't need it to work as before hook. Just `let`

route:
Never mind, I'll quick fix it! Thanks for your contribution!

PR #5, opened 2019-09-05
Fix a (presumable) typo around PDF printing.

nathanborgo:
The functionality saves a PDF to disk, but the comment suggests it would save a PNG.


route:
Thank you, typos)

PR #9, opened 2019-09-12
Allow to return pdf as binary

stoivo:
It is usefull when you want to store it to s3 or returning it in a request.

In reply to a request

```ruby
pdf_base64 = browser.pdf
pdf_binary = Base64.decode64(pdf_base64)
browser.quit

send_data pdf_binary, filename: "myfile.pdf", type: "application/pdf", disposition: "attachment"
```

OR

ActiveStorage S3

```ruby
pdf_base64 = browser.pdf
pdf_binary = Base64.decode64(pdf_base64)
browser.quit

invoice.pdf.attach(
  io: pdf_binary,
  filename: filename,
  content_type: "application/pdf",
)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -286,8 +286,8 @@ browser.screenshot(full: true, quality: 60) # "iVBORw0KGgoAAAANSUhEUgAABAAAAAMAC
 Saves PDF on a disk or returns it as base64.
 
 * options `Hash`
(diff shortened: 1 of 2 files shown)


stoivo:
Correct I dont want it too save to disk.

route:
I'll fix a bit your PR and it will work

route:
Fixed

stoivo:
Thanks

PR #12, opened 2019-09-17
Add paper format option to PDF

stoivo:
- rename option so all options is matching what the protocoll expect

### DIFF
--- a/README.md
+++ b/README.md
@@ -285,14 +285,14 @@ Saves PDF on a disk or returns it as base64.
   * :scale `Float` zoom in/out
   * :format `symbol` standard paper sizes :letter, :legal, :tabloid, :ledger, :A0, :A1, :A2, :A3, :A4, :A5, :A6
 
-  * :paperWidth `Float` set paper width
-  * :paperHeight `Float` set paper height
+  * :paper_width `Float` set paper width
+  * :paper_height `Float` set paper height
   * See other [native options](https://chromedevtools.github.io/devtools-protocol/tot/Page#method-printToPDF) you can pass
 
 ```ruby
 browser.goto("https://google.com/")
 # Save to disk as a PDF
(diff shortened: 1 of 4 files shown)


route:
This is all looks great except maybe camel-cased options. I know it's how they are represented in Chrome but in Ruby we don't use such, thus I decided to use underscore.

stoivo:
I agree that undesscore names are nices, and the ruby way. I would agrue that camed-case option is fine here since we are telling the user that we are passing it to [Page.printToPdf](https://chromedevtools.github.io/devtools-protocol/tot/Page#method-printToPDF) and you can add any options from that one. Maybe it is better to match the protocol 1-to-1 rather then maping everykey. I did not check what you did for other calls but I think is is important to be consistent with throught the gem.

Adding more more thoughts what do you think?

stoivo:
Do you have some secound thought?  or do you want me to implement all as snake case?

route:
Thanks @stoivo I didn’t use camel case anywhere in this gem or cuprite. I guess we better stick with ruby rather than chrome, because as a user I would expect it to be underscored and chrome internal calls are usually hidden from end user.

stoivo:
Right. Then I will adopt how you would like to have it.

These are the keys today 

 * landscape
 * displayHeaderFooter
 * printBackground
 * scale
 * paperWidth
 * paperWidth
 * marginTop
 * marginBottom
 * marginLeft
 * marginRight
 * pageRanges
 * ignoreInvalidPageRanges
 * headerTemplate
 * footerTemplate
 * preferCSSPageSize
 * transferMode

How would you like to do the mapping? convert case and a hash with exceptions

route:
Yea let's do the mapping manually like it was done in the method or if it's too many options we may copy-paste camelize method from `AS`, don't want to add such a huge dependency for just one method.

route:
@stoivo sorry I didn't notice when you force pushed, I'll check soon

route:
Thank you!

PR #16, opened 2019-09-25
Remove temporary directory when done

sigmike:
(imported from https://github.com/machinio/cuprite/pull/89 with a few changes)

The `user-data-dir` created [here](https://github.com/route/ferrum/blob/120dfe721b516a43db5b29a97cdff4b9e6bb0169/lib/ferrum/browser/process.rb#L122) is never removed. Ruby doesn't delete paths created with `Dir.mktmpdir` unless a block is provided[[1]](https://ruby-doc.org/stdlib-2.6.3/libdoc/tmpdir/rdoc/Dir.html#method-c-mktmpdir).

These changes remove the temporary directory either during the call to `stop` or when the process instance is garbage collected, inspired by what's done to kill the process.

Note that if it's garbage collected the directory is removed just before the Chrome process is killed.

It may be safer to have a single finalizer that does both the killing and the removing. It would be removed and defined again with new arguments when something changes (after the process is spawned for example).

### DIFF
--- a/lib/ferrum/browser/process.rb
+++ b/lib/ferrum/browser/process.rb
@@ -84,6 +84,15 @@ module Ferrum
         end
       end
 
+      def self.directory_remover(path)
+        proc do
+          begin
+            FileUtils.remove_entry(path)
+          rescue Errno::ENOENT
+          end
+        end
+      end
+
       def self.detect_browser_path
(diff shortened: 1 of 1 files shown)


route:
Thank you @sigmike! I'm fine with this solution

PR #27, opened 2019-10-17
Remove deprecated `Network.continueInterceptedRequest` calls

MSNexploder:
Switch from deprecated `Network.continueInterceptedRequest` to `Fetch.continueRequest`.
This resulted in a small API change. Auth callbacks are now separated from regular request callbacks and registration for them are now handled separately.

Additionally I added the ability for custom request fulfilment but I'm not 100% sure about the API design.

### DIFF
--- a/README.md
+++ b/README.md
@@ -370,6 +370,8 @@ browser.network.intercept
 browser.on(:request) do |request|
   if request.match?(/bla-bla/)
     request.abort
+  elsif request.match?(/lorem/)
+    request.respond(body: "Lorem ipsum")
   else
     request.continue
   end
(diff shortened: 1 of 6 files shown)


route:
Oh wow, this is great! Thank you! My main concern is that `Fetch` appeared not so long ago, what would happen if user would use outdated browser, which is usually the case for CI and linux packages?

MSNexploder:
Right now intercepting requests (including auth handling) on outdated browsers would stop working. According to [ChromeDevTools](https://github.com/ChromeDevTools/devtools-protocol/blob/master/changelog.md#2018-10-31-121551--diff-9801545fe05597) fetch was added almost exactly one year ago. Which feels like it *should* be long enough.

For example `Page.captureSnapshot` was added shortly after `Fetch` and while not as fundamental as intercepting requests I assume it's widely used.

route:
Sounds good then, let's try to switch to fetch

route:
Thank you! Didn't find any issues with cuprite integration

Issue #18, opened 2019-09-30
Full screenshots don't show page content


Odaeus:
Hi, I'm finding Ferrum to be a great lib, thank you! I have a bug to report when taking screenshots.

The `full: true` option when passed to `#screenshot` does not show the full page, instead it only renders the visible area of the page according to the viewport and then the rest is a solid background colour. I think the tests don't catch this because they only check the image dimensions.

Here's an example:
```ruby
b = Ferrum::Browser.new
b.goto "http://www.csszengarden.com/"
b.resize width: 800, height: 200
b.screenshot(path: "zg200.png", full: true)
```
<img src="https://user-images.githubusercontent.com/2150/65871977-d8d99280-e37f-11e9-87a8-34d68c02b65d.png" width=200>

```ruby
b.resize width: 800, height: 600
b.screenshot(path: "zg600.png", full: true)
```
<img src="https://user-images.githubusercontent.com/2150/65872052-00305f80-e380-11e9-8a0a-139b6afd351f.png" width=200 />

```ruby
b.resize width: 800, height: b.evaluate("document.documentElement.offsetHeight")
b.screenshot(path: "zgfull.png", full: true)
```
<img src="https://user-images.githubusercontent.com/2150/65872183-51405380-e380-11e9-94b8-825adf7c8e58.png" width=200 />

route:
Thanks for reporting, it's great that it's a public website I'll take a look.

Mifrill:
```
it "issue #18" do
	browser.goto "http://www.csszengarden.com/"
	sleep 1
	browser.resize width: 800, height: 200
	browser.screenshot(path: file, full: true)
	expect(File.exist?(file)).to be true
	expect(browser.viewport_size).to eq([800, 200])
end
```
Result in https://github.com/Mifrill/ferrum/tree/issue%2318_full-screenshots-dont-show-page-content
![screenshot](https://user-images.githubusercontent.com/23525618/67149125-e5874180-f2c0-11e9-9678-632d058da45e.png)

PR #29, opened 2019-10-19
Issue#18 full screenshots dont show page content

Mifrill:
Closes: #18 

According to [this WiKi](https://github.com/cyrus-and/chrome-remote-interface/wiki/Take-page-screenshot) we should override the metrics before try of taking a full-page screenshot and also better to revert them after the capture action to not confuse the data for the current instance of `@client`.


route:
Thanks!

Mifrill:
@route Thanks to you! This is a fantastic gem! :100:

PR #33, opened 2019-11-04
Use default user-agent when overriding special headers.

Slotos:
Accept-Language is overridden through [`setUserAgentOverride` method in Network domain](https://chromedevtools.github.io/devtools-protocol/tot/Network#method-setUserAgentOverride), which declares User-Agent to be a required parameter. As we don't always care to set User-Agent
at the same time as we set Accept-Language, we should store chrome own user agent as reported on `/json/version` endpoint, and reuse it if needed.

### Issue reproduction path

```
require 'ferrum'
browser = Ferrum::Browser.new
browser.headers.add({"Accept-Language"=>"en"}) # => Ferrum::BrowserError: Invalid parameters
```

### Expected outcome, as implemented

Provided headers are added.
User-Agent defaults to chrome default one, unless otherwise specified.

### Actual outcome, as exhibited by master

`Ferrum::BrowserError: Invalid parameters` error is raised

### DIFF
--- a/lib/ferrum/browser.rb
+++ b/lib/ferrum/browser.rb
@@ -25,6 +25,7 @@ module Ferrum
                 evaluate evaluate_on evaluate_async execute
                 add_script_tag add_style_tag
                 on] => :page
+    delegate %i[default_user_agent] => :process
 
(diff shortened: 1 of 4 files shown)


route:
Thank you! Awesome

PR #46, opened 2020-02-05
Raise a specific error when browser takes too long to start

nenoganchev:
A generic `RuntimeError` was raised when the browser failed to respond with a WebSocket URL on time. This makes it difficult for client code to distinguish and inconvenient to handle this particular error.

We do need to handle it in our code, since we do encounter it – very, very rarely, but it does occur occasionally. To that end, this PR adds a new error type for timeout errors in `Ferrum::Browser::Process` and raises it when the browser fails to respond with a WebSocket URL on time.

### DIFF
--- a/lib/ferrum.rb
+++ b/lib/ferrum.rb
@@ -30,6 +30,12 @@ module Ferrum
     end
   end
 
+  class ProcessTimeoutError < Error
+    def initialize(message = "Timed out waiting for browser process to start")
+      super
+    end
+  end
+
   class DeadBrowserError < Error
     def initialize(message = "Browser is dead or given window is closed")
       super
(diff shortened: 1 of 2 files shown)


route:
@nenoganchev thank you so much for your PRs both look great to me, though I think with the last Chrome update tests are failing. I'll fix them and merge then.

nenoganchev:
@route you are welcome, I'm glad I could contribute back to your work. Working with your gem has been a pleasure so far, so thank you for making it.

I noticed the failing test and tried to reproduce it locally, but on my machine another set of tests were failing (both in my branches and on master), so I decided I may not have the proper environment and gave up. Let me know if I can be of any help.

Cheers.

route:
Chrome 80 introduced new feature. When we select text with ctrl+shift+left earlier versions of Chrome would select the whole word, but 80 selects only exclamation mark in the word "Hello!". I'm merging your PRs

PR #51, opened 2020-02-21
Use scrollHeight & scrollWidth for fullscreen screenshot

### LINKED ISSUE #49: Use scrollHeight/Width instead of offset for full?
The `full` option for screenshots is currently using the `offsetHeight` and `offsetWidth`:

https://github.com/route/ferrum/blob/0e96c1cf87ac768df521a6f321ced22f13899b7e/lib/ferrum/page/screenshot.rb#L49-L54

However, on my device™ (in Chrome Version 80.0.3987.116) this returns the size of the viewport. Thus when I run `browser.screenshot(full: true, "google.png")` I get the same result as without the `full: true` option. 

Using the `scrollHeight`/`scrollWidth` instead returns the actual height/width of the rendered site.

PS: thank you for this gem, it has been a joy to use :)

coffeejunk:
If the `html` tag has the `height` or `width` set to `100%`, `document.documentElement.offsetHeight` (`offsetWidth`) as well as `document.documentElement.clientHeight` (`clientWdith`) return the
height / width of the *viewport*.

https://github.com/coffeejunk/ferrum/blob/6906d98f48042a21c49ccccf3cf15c89541b0bc0/spec/support/views/custom_html_size.erb#L1-L19

Screenshot from 2020-02-21 16-41-39

Using `document.documentElement.scrollHeight` (`scrollWidth`) fixes this behavior as it returns the full height/width of the rendered page.

> The scrollHeight value is equal to the minimum height the element
> would require in order to fit all the content in the viewport without
> using a vertical scrollbar.
https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollHeight

> The scrollWidth value is equal to the minimum width the element would
> require in order to fit all the content in the viewport without using
> a horizontal scrollbar.
https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollWidth

Fixes #49

### DIFF
--- a/spec/support/views/custom_html_size.erb
+++ b/spec/support/views/custom_html_size.erb
@@ -1,13 +1,19 @@
 <html>
-<head>
-  <title>custom_size</title>
-  <style>
-    html {
-      width: 1280px;
-      height: 1024px;
-    }
-  </style>
-</head>
+  <head>
+    <title>custom_size</title>
+    <style>
+      html {
+        width: 100%;
+        height: 100%;
+      }
+
+      .spacer {
+        width:  1272px; /* using scrollWidth  adds 8px */
+        height: 1008px; /* using scrollHeight adds 16px */
+      }
+    </style>
+  </head>
   <body>
+    <div class="spacer"></div>
   </body>
 </html>
--- a/lib/ferrum/page/screenshot.rb
+++ b/lib/ferrum/page/screenshot.rb
@@ -48,8 +48,8 @@ module Ferrum
 
       def document_size
         evaluate <<~JS
-          [document.documentElement.offsetWidth,
-           document.documentElement.offsetHeight]
+          [document.documentElement.scrollWidth,
+           document.documentElement.scrollHeight]
         JS
       end
 


route:
Looks good! I need to run Cuprite tests with this change as well. Thank you!

spec/support/views/custom_html_size.erb
```diff
@@ -1,13 +1,19 @@
...
+    <title>custom_size</title>
+    <style>
+      html {
+        width: 100%;
+        height: 100%;
+      }
+
+      .spacer {
+        width:  1272px; /* using scrollWidth  adds 8px */
+        height: 1008px; /* using scrollHeight adds 16px */
```
Mifrill, line 12: @coffeejunk I think, I got why we need to use `scroll` references over `offset`, but why is it different from `Width` gap, it should be the same, isn't'?
`height: 1016px; /* using scrollHeight adds 8px */`

route:
Specs in master are now only passing with these changes https://github.com/rubycdp/ferrum/pull/138

spec/support/views/custom_html_size.erb
coffeejunk: @Mifrill to be quite honest, the change is almost a year old, and I don't remember anymore why it would've been 16px vs. 8px. From what I see right now, I think you are right in that it should be 8px. :/

Mifrill: @coffeejunk no problem, yeah, this "gap" by the defalut Chrome margins:
https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/core/html/resources/html.css;l=57-60

PR #66, opened 2020-04-14
Add Xvfb support

sebyx07:
I found using Xvfb useful, some sites detect chrome headless.

With this, there is no need for the gem user to manage Xvfb himself. the `DISPLAY` number is generated using the current time in nanoseconds % 100.000.000 so it's a very low chance of colliding

### DIFF
--- a/lib/ferrum/browser/process.rb
+++ b/lib/ferrum/browser/process.rb
@@ -19,7 +19,7 @@ module Ferrum
 
       attr_reader :host, :port, :ws_url, :pid, :command,
                   :default_user_agent, :browser_version, :protocol_version,
-                  :v8_version, :webkit_version
+                  :v8_version, :webkit_version, :environment
 
 
       extend Forwardable
@@ -63,6 +63,7 @@ module Ferrum
           return
         end
 
+        @environment = Ferrum::Browser::Environment.new(options)
         @logger = options[:logger]
         @process_timeout = options.fetch(:process_timeout, PROCESS_TIMEOUT)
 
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## [0.9.0] - (Apr 14, 2020) ##
```
Mifrill, line 1: @sebyx07 you need add this line:
```
[0.9.0]: https://github.com/rubycdp/ferrum/compare/v0.8.0...v0.9.0
```
[here](https://github.com/sebyx07/ferrum/blob/master/CHANGELOG.md#L190)  to make link works

lib/ferrum/version.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 module Ferrum
-  VERSION = "0.8"
+  VERSION = "0.9"
```
route, line 4: Let's keep the version, I didn't plan to make a release that fast

CHANGELOG.md
route, line 1: Let's make it unreleased , I'll bump when I release new version

lib/ferrum/browser/process.rb
```diff
@@ -63,6 +63,7 @@ def initialize(options)
           return
         end
 
+        @environment = Ferrum::Browser::Environment.new(options)
```
route, line 66: You can use just `Environment`

route:
I changed the implementation in master, it now accepts xvfb along with headless option

PR #70, opened 2020-04-19
issue#67 Manage of 'disable-web-security' browser flag as option

### LINKED ISSUE #67: Remove browser_options in DEFAULT_OPTIONS
Currently the `DEFAULT_OPTIONS` in `lib/ferrum/browser/chrome.rb` can't be removed by the options given to `Ferrum::Browser.new`. Is there any way to do that?

The problem I'm facing is that some website check the "origin" HTTP header on the server side, and with "--disable-web-security" Chrome doesn't send the origin header at all.

Workaround I'm using now is:
```ruby
opt = Ferrum::Browser::Chrome.send(:remove_const, :DEFAULT_OPTIONS).dup
opt.delete('disable-web-security')
Ferrum::Browser::Chrome.const_set(:DEFAULT_OPTIONS, opt)
```

Mifrill:
closes #67

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,7 +175,7 @@ Ferrum::Browser.new(options)
   * `:ws_max_receive_size` (Integer) - How big messages to accept from Chrome
       over the web socket, in bytes. Defaults to 64MB. Incoming messages larger
       than this will cause a `Ferrum::DeadBrowserError`.
-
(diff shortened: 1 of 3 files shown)


route:
I think that would make it very complicated to implement for every single options, I like your second approach

PR #77, opened 2020-04-27
Fix main_frame for target _blank links

betelgeuse:
When opening a target _blank link to a page with multiple iframes
the main_frame tracking code got it wrong and picked one of the
iframes as the main_frame. In this case Runtime.executionContextCreated
gets tricked last for the root frame.

For this fix I took the must straightforward route to fixing the problem. However, what is the reason  that Ferrum is doing all this complex tracking of frames when it could just ask Chrome using the API Page.getFrameTree?

### DIFF
--- a/lib/ferrum/page/frames.rb
+++ b/lib/ferrum/page/frames.rb
@@ -75,8 +75,11 @@ module Ferrum
           frame_id = params.dig("context", "auxData", "frameId")
 
           unless @main_frame.id
-            @main_frame.id = frame_id
-            @frames[frame_id] = @main_frame
+            root_frame = command('Page.getFrameTree').dig('frameTree', 'frame', 'id')
+            if frame_id == root_frame
+              @main_frame.id = frame_id
+              @frames[frame_id] = @main_frame
+            end
           end
 
           frame = @frames[frame_id] || Frame.new(frame_id, self)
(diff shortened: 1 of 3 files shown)


route:
Chrome streams events on frames while `Page.getFrameTree` should be called manually I guess that was the reason. It should be a combination of both I think.

betelgeuse:
> Chrome streams events on frames while `Page.getFrameTree` should be called manually I guess that was the reason. It should be a combination of both I think.

It currently is in the pull request. However, this doesn't really explain why complexity over calling Page.getFrameTree whenever the information is needed. It seems the two approaches would provide the same functionality but the latter is simpler and as such less prone to bugs. Premature optimisation in play?

betelgeuse:
@route would be nice to get this bug fix moving forward

Mifrill (approved):
LGTM!

Mifrill (approved):
LGTM!

lib/ferrum/page/frames.rb
```diff
@@ -75,8 +75,11 @@ def frames_subscribe
           frame_id = params.dig("context", "auxData", "frameId")
 
           unless @main_frame.id
-            @main_frame.id = frame_id
-            @frames[frame_id] = @main_frame
+            root_frame = command('Page.getFrameTree').dig('frameTree', 'frame', 'id')
```
Mifrill, line 78: I think it's nice to have `public` method for this `Ferrum::Page::Frames#root_frame`

```
def root_frame
  command('Page.getFrameTree').dig('frameTree', 'frame', 'id')
end
```

PR #79, opened 2020-05-15
Update example Google form selector

ttilberg:
The current xpath selector is not valid (at least not here, now).

This version of the selector is more ambiguous, targeting only the text input, and uses the `q` param.


ttilberg:
In retrospect, perhaps `at_xpath` was being used just to demonstrate that it exists, since `at_css` is used shortly after. I'll change that.

route:
Thank you! Yes it was used just to show that it exists :)

PR #80, opened 2020-05-19
Relax addressable version

azranel:
# Description

I'm thinking about using `ferrum` in my application but due to some other dependencies, I'm not able due to what `addressable` version is specified in `ferrum`. I checked the [addressable changelog](https://github.com/sporkmonger/addressable/blob/master/CHANGELOG.md#addressable-250) and it seems that there weren't that many changes between addressable 2.5 and addresssable 2.6 so I think we could relax this version a little bit.


route:
Thanks, sounds good!

PR #87, opened 2020-06-09
Fix Ferrum::Browser::Subscriber thread safety

betelgeuse:
The on method is called without going through async so all
the data structures need to be thread safe.

https://ruby-concurrency.github.io/concurrent-ruby/1.1.6/Concurrent/Async.html

Changing the call sites would be more in line with Async
documentation but that would be a change in behavior.

### DIFF
--- a/lib/ferrum/browser/subscriber.rb
+++ b/lib/ferrum/browser/subscriber.rb
@@ -13,7 +13,7 @@ module Ferrum
 
       def initialize
         super
-        @on = Hash.new { |h, k| h[k] = [] }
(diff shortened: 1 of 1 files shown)


betelgeuse:
This still leaves the issue that between the call for size here and enumeration the collection could be modified.

https://github.com/rubycdp/ferrum/blob/19bf88b8873d4507d645e51e964c4482f8f4a1cc/lib/ferrum/browser/subscriber.rb#L26

route:
Thanks

Issue #78, opened 2020-05-08
Chromium failing to start with "failed to dup supplied urandom fd: Bad file descriptor"


jarthod:
Hello,

I was seing this error randomly in my builds:
```
Browser did not produce websocket url within 5 seconds
```
The machine was not saturated, of course I tried increasing it (to 20 seconds) but the problem was still the same so it was clear it was not a timing issue.

I tried to enable debug logging but it was way too verbose so impossible to find the error in here. I ended up looking at the code adding just a puts to see the output where the ws URL is missing:
```ruby
        unless ws_url
          @logger.puts(output) if @logger
          puts output
          raise ProcessTimeoutError.new(timeout)
        end
```
And so when the errors occurs I got this:
```
failed to dup supplied urandom fd: Bad file descriptor
```

This error comes from chromium code I guess, but I'm not sure why I have it here and not when I use chromium directly. It doesn't look like Ferrum is "supplying" a custom FD for this.
```
> chromium --version
Chromium 81.0.4044.113 snap
```

I tried starting chromium manually with the exact same parameters:
```
/snap/bin/chromium --headless --disable-gpu --hide-scrollbars --mute-audio --enable-automation --disable-web-security --disable-session-crashed-bubble --disable-breakpad --disable-sync --no-first-run --use-mock-keychain --keep-alive-for-test --disable-popup-blocking --disable-extensions --disable-hang-monitor --disable-features=site-per-process,TranslateUI --disable-translate --disable-background-networking --enable-features=NetworkService,NetworkServiceInProcess --disable-background-timer-throttling --disable-backgrounding-occluded-windows --disable-client-side-phishing-detection --disable-default-apps --disable-dev-shm-usage --disable-ipc-flooding-protection --disable-prompt-on-repost --disable-renderer-backgrounding --force-color-profile=srgb --metrics-recording-only --safebrowsing-disable-auto-update --password-store=basic --window-size=1024,768 --remote-debugging-port=0 --remote-debugging-address=127.0.0.1 --user-data-dir=/tmp/d20200508-7638-17rc0fb
```
Many times but never reproduced this error message.

I run my tests in parallel using `paraspec`, I could reproduce the problem even with a very low concurrency of 2 (on a 6 cores CPU) so the problem is not starting time or slowdown it seems, but maybe concurrency issue somewhere?

I managed to write a small Ferrum script to reproduce the issue using fork:
```ruby
#!/usr/bin/env ruby

require 'ferrum'

10.times do
[...]

PR #92, opened 2020-07-02
Include process output in ProcessTimeoutError

odlp:
Hello,

First off, thanks for this great library and Cuprite too :raised_hands:

I've experienced a similar issue to #78, in my case relating to Chrome permissions in Docker and found it hard to diagnose the root cause. Although the logging mechanism is useful, I would like to propose including the `output` captured on the `ProcessTimeoutError` itself.

This allows the integrator to capture the exception, which personally I find handy for debugging purposes:

```ruby
begin
  # Something with Ferrum
rescue ProcessTimeoutError => ex
  puts ex.output # or binding.pry etc
end
```

Many thanks,
Oli

### DIFF
--- a/lib/ferrum.rb
+++ b/lib/ferrum.rb
@@ -37,7 +37,10 @@ module Ferrum
   end
 
   class ProcessTimeoutError < Error
-    def initialize(timeout)
+    attr_reader :output
+
+    def initialize(timeout, output)
+      @output = output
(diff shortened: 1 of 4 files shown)


Mifrill (approved):
:+1:

route:
Thank you @odlp for the PR and sorry it took me so long to review it and merge, but I'm trying to improve that part and there's a small team behind RubyCDP now. Thanks @Mifrill

Issue #89, opened 2020-06-23
Warnings


jamarat:
There are a two warnings when running `rake test` for app build with ferrum:
```
/Library/Ruby/Gems/2.6.0/gems/ferrum-0.8/lib/ferrum/context.rb:45: warning: assigned but unused variable - target_id
/Library/Ruby/Gems/2.6.0/gems/ferrum-0.8/lib/ferrum/browser/process.rb:96: warning: instance variable @user_data_dir not initialized
```
Probably nothing dangerous but a bit unclean.

Mifrill:
@jamarat thanks for the raise the flag!

The first case looks weird:


https://github.com/Mifrill/ferrum/blob/master/lib/ferrum/context.rb#L44-L52
```
    def create_target
      target_id = @browser.command("Target.createTarget",
                                   browserContextId: @id,
                                   url: "about:blank")["targetId"]
      target = @pendings.take(@browser.timeout)
      raise NoSuchTargetError unless target.is_a?(Target)
      @targets[target.id] = target
      target
    end
```
we probably should remove the declaration of `target_id` at all here:
```
	  target = @pendings.take(@browser.timeout)
      raise NoSuchTargetError unless target.is_a?(Target)
      @targets[target.id] = target
      target
```
or we should to apply something like this:
```
target = @targets[target_id] || @pendings.take(@browser.timeout)
```
because we may have nothing in `@pendings` `Concurrent::MVar` on `@pendings.take` fire, according to logic in `#add_target`:
```
	def add_target(params)
      target = Target.new(@browser, params)
      if target.window?
        @targets[target.id] = target
      else
        @pendings.put(target, @browser.timeout)
      end
    end
```

route:
@Mifrill I think we can remove it, agree

Mifrill:
@route we can't just remove this declaration, because we should trigger the `subscriber` event here:
https://github.com/Mifrill/ferrum/blob/ecab9ec5a0aa8454e7021bc5fcfc8c5f1c3fb6f7/lib/ferrum/contexts.rb#L48 
by `browser.client.on("Target.targetCreated")` to `Creates a new page.`
https://chromedevtools.github.io/devtools-protocol/tot/Target/#method-createTarget

PR #100, opened 2020-07-27
Issue#89 resolve warnings

Mifrill:
Closes: #89

### DIFF
--- a/lib/ferrum/browser/process.rb
+++ b/lib/ferrum/browser/process.rb
@@ -64,6 +64,7 @@ module Ferrum
           return
         end
 
+        @user_data_dir = nil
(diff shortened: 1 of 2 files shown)


lib/ferrum/browser/process.rb
```diff
@@ -115,13 +115,15 @@ def restart
 
       private
 
+      attr_accessor :user_data_dir
```
route, line 118: We should just put `@user_data_dir = nil` to initializer, because there's no sense in attr_accessor

Mifrill: okay, thanks

route:
Thank you!

PR #111, opened 2020-09-03
Rename temporary user data directory

sigmike:
The default name prefix for the temporary user data dir is `d` which is
not very helpful when trying to figure out where the directory comes
from. The patch makes the directory name more explicit.


route:
Thanks, looks good :+1:

PR #116, opened 2020-09-11
WindowBounds have not only width and height but also left and top

Nakilon:
The test does not pass in headless -- `get_position` returns the same after repositioning, I don't know why.


lib/ferrum/page.rb
```diff
@@ -89,15 +89,14 @@ def close
     end
 
     def resize(width: nil, height: nil, fullscreen: false)
-      result = @browser.command("Browser.getWindowForTarget", targetId: @target_id)
-      @window_id, @bounds = result.values_at("windowId", "bounds")
+      @window_id = @browser.command("Browser.getWindowForTarget", targetId: @target_id)["windowId"]
```
Mifrill, line 92: it makes sense, due to we really didn't use `@bounds`

lib/ferrum/page.rb
```diff
@@ -89,15 +89,14 @@ def close
...
 
       if fullscreen
         width, height = document_size
         @browser.command("Browser.setWindowBounds", windowId: @window_id, bounds: { windowState: "fullscreen" })
       else
         @browser.command("Browser.setWindowBounds", windowId: @window_id, bounds: { windowState: "normal" })
-        @browser.command("Browser.setWindowBounds", windowId: @window_id, bounds: { width: width, height: height, windowState: "normal" })
+        @browser.command("Browser.setWindowBounds", windowId: @window_id, bounds: { width: width, height: height })
```
Mifrill, line 99: it makes sense too, coz `setWindowBounds` `Leaves unspecified fields unchanged.`:
https://chromedevtools.github.io/devtools-protocol/tot/Browser/#method-setWindowBounds

spec/browser_spec.rb
```diff
@@ -87,11 +87,19 @@ module Ferrum
       expect(browser.viewport_size).to eq([200, 400])
     end
 
-    it "allows the viewport to be resized by fullscreen" do
+    it "allows the viewport to be resized to fullscreen and back" do
```
Mifrill, line 90: I do believe it should be in a dedicated test, like: `resizes to "normal" from "fullscreen" window state`

spec/browser_spec.rb
```diff
@@ -87,11 +87,19 @@ module Ferrum
...
       browser.goto("/ferrum/custom_html_size")
       expect(browser.viewport_size).to eq([1024, 768])
       browser.resize(fullscreen: true)
       expect(browser.viewport_size).to eq([1280, 1024])
+      browser.resize(width: 200, height: 400)
+      expect(browser.viewport_size).to eq([200, 400])
```
Mifrill, line 96: this test makes sense, to prevent this error:
```
Ferrum::BrowserError:
       To resize minimized/maximized/fullscreen window, restore it to normal state first.
```
let's split into dedicated test:
```
      it "resizes to 'normal' from 'fullscreen' window state" do
        browser.go_to(path)
        browser.resize(fullscreen: true)
        browser.resize(width: 200, height: 400)
        expect(browser.viewport_size).to eq([200, 400])
      end
```

Mifrill (changes requested):
@Nakilon thanks for the PR! I see 2 logic-branches here:
1 - refactor `resize` method and add missed test for error:
`To resize minimized/maximized/fullscreen window, restore it to normal state first`
2 - `get_position`/`set_position` implementation
So, let's split it for 2 PR's

A bit refactor of `Page#resize` like this:
```
    def resize(width: nil, height: nil, fullscreen: false)
      if fullscreen
        width, height = document_size
        set_window_bounds(windowState: "fullscreen")
      else
        set_window_bounds(windowState: "normal")
        set_window_bounds(width: width, height: height)
      end

      command("Emulation.setDeviceMetricsOverride", slowmoable: true,
                                                    width: width,
                                                    height: height,
                                                    deviceScaleFactor: 1,
                                                    mobile: false,
                                                    fitWindow: false)
    end

    def window_id
      @browser.command("Browser.getWindowForTarget", targetId: @target_id)["windowId"]
    end

    def set_window_bounds(bounds = {})
      @browser.command("Browser.setWindowBounds", windowId: window_id, bounds: bounds)
    end
```
And keep this PR for `get_position`/`set_position` implementation.

So, please merge it with the updated master.

About:
> The test does not pass in headless

Can you put some more details? It would be nice to add failed test for this.

spec/browser_spec.rb
Nakilon: I don't remember for sure but maybe such test would need to switch from windowed to fullscreen first? Then it would repeat "windowed to fullscreen".

Mifrill: it's okay to "repeat" coz, we have 2 different cases:
1 - resize with fullscreen mode -> viewport_size depends on document sizes
2 - resize from fullscreen mode to normal -> viewport_size depends on specified arguments

Mifrill:
@Nakilon FYI: I've moved the refactoring of `Page#resize` into the dedicated PR: https://github.com/rubycdp/ferrum/pull/152
Let's keep only `get_position/set_position` implementation here and please share some details about the issue you mentioned.

Mifrill:
@Nakilon 
> The test does not pass in headless -- get_position returns the same after repositioning, I don't know why

This test is green for me locally, and the `set/get-position` working like a charm, can you please shed some light on your comment?

Nakilon:
> @Nakilon
> 
> > The test does not pass in headless -- get_position returns the same after repositioning, I don't know why
> 
> This test is green for me locally, and the `set/get-position` working like a charm, can you please shed some light on your comment?

I guess this line https://github.com/rubycdp/ferrum/pull/116/files#diff-e180ec3f0f9f10ea0a9dc5fe5e4f00f83ba5e263031a80d8954cc9bc52a14b0eR102 was failing unless I use the `headless: false`. So it kinda was able to move only that window that is visible. Was failing on macOS and I guess in CI too when the pull request didn't have merge conflict.

lib/ferrum/page.rb
```diff
@@ -108,6 +108,16 @@ def resize(width: nil, height: nil, fullscreen: false)
...
 
+    def get_position
+      @window_id = @browser.command("Browser.getWindowForTarget", targetId: @target_id)["windowId"]
+      @browser.command("Browser.getWindowBounds", windowId: @window_id).fetch("bounds").values_at("left", "top")
+    end
+
+    def set_position(left, top)
+      @window_id = @browser.command("Browser.getWindowForTarget", targetId: @target_id)["windowId"]
+      @browser.command("Browser.setWindowBounds", windowId: @window_id, bounds: { left: left, top: top })
+    end
```
Mifrill, line 119:
```suggestion
    def get_position
      @browser.command("Browser.getWindowBounds", windowId: window_id).fetch("bounds").values_at("left", "top")
    end

    def set_position(left, top)
      @browser.command("Browser.setWindowBounds", windowId: window_id, bounds: { left: left, top: top })
    end
```

lib/ferrum/browser.rb
```diff
@@ -109,6 +109,14 @@ def resize(**options)
...
     end
 
+    def get_position
+      page.get_position
+    end
+
+    def set_position(left, top)
+      page.set_position(left, top)
+    end
+
```
Mifrill, line 119: it can be delegated to page here:
https://github.com/rubycdp/ferrum/pull/116/files#diff-5af714046589275108757b6cc194e7ae6e35bc6bcdb723ecfa48aad1cf7232f0R20
```suggestion
    delegate :get_position, :set_position, to: :page

```

lib/ferrum/page.rb
Mifrill: now here we can use `window_id` method:
https://github.com/rubycdp/ferrum/pull/116/files#diff-2f9537c613f48bed5ff6ca273e2dac28886a32fe78dcd2dc7c7c798c891434d3L140-L142

Mifrill (approved):

route:
Looking good, let's add docs to readme as well

route (approved):

Issue #82, opened 2020-05-20
Implement wait_for_selector


route:
https://github.com/puppeteer/puppeteer/blob/master/docs/api.md#pagewaitforselectorselector-options

Daniel-ltw:
```ruby
      sleep(@browser.slowmo) if slowmoable && @browser.slowmo > 0
```

Is this part of addressing this? 
Was thinking of just doing a sleep on slowmo define time, we can just do an interval loop to wait on a specific element on the page to show. 

Because of this sleep, my current test run seems to take too long to complete.

Daniel-ltw:
This is my hacky attempt in my own spec_helper.rb. Rather than just using slowmo. 

```ruby
def wait_for(css_element, visibility: :all)
  @staled_retries ||= 120
  element = first(css_element, visible: visibility, wait: 0.3)
  warn_log "Found #{css_element}"
  yield(element) if block_given?
  element
rescue Capybara::ElementNotFound, Capybara::ExpectationNotMet => e
  warn_log "#{css_element} stale for: #{@staled_retries}"
  if @staled_retries.zero?
    page_path = save_page
    screenshot_path = save_screenshot
    debug_log "Page at #{page_path}"
    debug_log "Screenshot at #{screenshot_path}"
    raise e
  else
    @staled_retries -= 1
    sleep(0.2)
    retry
  end
end
```

shreeve:
I've been using this... seems like something like this should be in Ferrum.

```ruby
module Ferrum
  class Frame
    module DOM

      #
      # Finds a node using XPath or a CSS path selector, with waiting.
      #
      # @param [String] selector
      #   The XPath or CSS path selector.
      #
      # @param [Integer] init
      #   How long we should wait before starting to look.
      #
      # @param [Integer] wait
      #   How long we should wait for node to appear.
      #
      # @param [Integer] step
      #   How long to wait between checking.
      #
      # @return [Node, nil]
      #   The matching node.
      #
      # @example
      #   browser.with("a[aria-label='CA']", init:1, wait:5, step:0.2) # => Node
      #
      def with(selector, init:nil, wait:1, step:0.1)
        sleep(init) if init
        meth = selector.start_with?("/") ? :at_xpath : :at_css
        until node = send(meth, selector)
          (wait -= step) > 0 ? sleep(step) : break
        end
        node
      end
    end
  end
end
```

I use it like this:

```ruby
[...]

Issue #81, opened 2020-05-20
Wait for the appearance of elements


alexanderadam:
Hi folks,

I love ferrum but is there any inbuilt possibility to wait for the appearance of an element?
Or is something like this at least planned?
Or is it easily possible to integrate something like [`wait-for-element.js`](https://github.com/azu/wait-for-element.js)?

route:
@alexanderadam if you are using it for tests you need [Cuprite](https://github.com/rubycdp/cuprite) you can also use it for the same purpose as Ferrum with slightly different API and Capybara with `synchronize` support where it tries to find element during some period of time. As for Ferrum `wait_for_selector` is going to be implemented in future versions.

alexanderadam:
> if you are using it for tests […]

I'm not.

> `wait_for_selector` is going to be implemented in future versions

Sounds great! But can we leave the issue open then until it is fixed? This way people can subscribe and follow it's status.

route:
I've created https://github.com/rubycdp/ferrum/issues/82

alexanderadam:
Ah okay, thank you. It wasn't there when I opened this issue though. :wink:

PR #118, opened 2020-10-07
fix #95: error 'Could not compute content quads'

### LINKED ISSUE #95: How to handle or circumvent Ferrum::BrowserError "Could not compute content quads."?
I don't find anything regarding this.

alexanderadam:
Hi @route,

sorry that it took so long. :disappointed: 

So the thing is: in _theory_ [`DOM.getContentQuads`](https://github.com/rubycdp/ferrum/blob/6d1612de80128e4dd20a6cd0881c2ce64bc21d1e/lib/ferrum/node.rb#L149) is absolutely the right thing to do. Depending on the used browser, this seem to be unstable.

Thus, if we're running into this unstable state (and only then!) we're going to calculate the position via JS.
This might be slower, but at least it's working flawlessly. And hopefully `getContentQuads` will become more stable and we won't run into this anymore anyway.

I hope it's working for you and others. At least it's working in the projects of my clients.

**PS:** If you're going to merge this: [would you be so kind and add the `hacktoberfest-accepted` label to this PR](https://hacktoberfest.digitalocean.com/hacktoberfest-update)? Thank you in advance!
**PPS:** We should definitely get rid of all the `sleep`s in the specs and migrate to a more stable solution. Maybe this would be related to #81 / #82 (I still don't know why you created another issue for this). :wink:

### DIFF
--- a/lib/ferrum/node.rb
+++ b/lib/ferrum/node.rb
@@ -141,9 +141,20 @@ module Ferrum
       end.map { |q| to_points(q) }.first
 
       get_position(points, x, y, position)
+    rescue Ferrum::BrowserError => e
+      return raise unless e.message&.include?('Could not compute content quads')
+
+      find_position_via_js
     end
 
     private
+
+    def find_position_via_js
+      [
+        evaluate('this.getBoundingClientRect().left + window.pageXOffset + (this.offsetWidth / 2)'), # x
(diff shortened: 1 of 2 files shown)


route:
Thank you

alexanderadam:
I have to thank _you_ buddy. :raised_hands: 
Thank you so much for the RubyCDP projects. I wish @crystal-lang would have libraries like these.

route:
Haha maybe some day :)

PR #119, opened 2020-10-14
Screenshot PNG with transparent background

ondrejbartas:
sometimes you need to have transparent background,

I took that command from puppeteer codebase. 

It is up to you if you want to somehow implement it.


route:
Sounds good thanks, if you put it to `screenshot` method as an option I would gladly accept a PR ;)

route:
@Mifrill this one should be easy to implement

Mifrill:
@route I think, it's better to allow set color by RBGA over than use transparent_background option only, due to the browser-specific, for example, Safari case:
> Safari. It’s more reliable to use rgba(255, 255, 255, 0) instead
https://stackoverflow.com/questions/46309641/what-happens-in-safari-with-the-transparent-color

Mifrill:
Apart from that: I'd like to ask how to test background color in the file? It looks like we need to add the external-gem dependency for this, like https://rmagick.github.io/imageattrs.html#background_color

route:
I think it's not Ferrum's responsibility to check that transparency works, we are good.

Issue #106, opened 2020-08-20
Add `go_to` alias for `goto`


AlexWayfer:
I don't know why there is `goto`. I didn't find such method in [the official documentation](https://chromedevtools.github.io/devtools-protocol/). But there are two words, and I think we should use underscore. For backward compatibility, I suggest not to rename, but add the alias. What do you think?

AlexWayfer:
> Sounds good!

OK, then I'm going to create a PR>

AlexWayfer:
But at first I want to done some documentation and formatting issues, like #107, #108, #108 + unready RuboCop addition.

AlexWayfer:
> But at first I want to done some documentation and formatting issues, like #107, #108, #108 + unready RuboCop addition.

Mentioned PRs are merged (thank you), now there is #110 for "RuboCop addition". Pretty doubtful, but I think useful.

AlexWayfer:
Is it can be closed or maybe we want to deprecate `goto` method?

PR was merged.

/cc @alexanderadam

AlexWayfer:
When we can expect a new release with this?

route:
I need to work on a few more features and then will release new version

trkoch:
`README` already suggests `go_to` since it's in master. Just wanted to mention I was tricked by this.

PR #120, opened 2020-10-26
chores #106: alias `go_to` to `goto` and use `go_to` within ferrum for Ruby consistency

alexanderadam:
Hi @route,

it's me again. :wave:
I thought that #106 makes sense and thus I gave it a go.
~~I used the inbuilt deprecation method to deprecate `goto` (since it's more Ruby like to use method names with underscore divided words) but I didn't know when you would like to deprecate it?~~
~~I've set it now to April 2021. I mean, of course it isn't a hard date but what date would make sense here?~~

~~This is how it looks like when you call the deprecated name:~~

NOTE: Ferrum::Page#goto is deprecated; use go_to instead. It will be removed on or after 2021-04-01. Ferrum::Page#goto called from /home/user/.rbenv/versions/2.7.1/lib/ruby/2.7.0/forwardable.rb:235.

**EDIT:** I removed the deprecation warning now, since deprecating the former name is obviously not wanted.

**PS:** If you're going to merge this: would you be so kind and add the [`hacktoberfest-accepted`](https://hacktoberfest.digitalocean.com/hacktoberfest-update) label to this PR? Thank you in advance!

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ Navigate to a website and save a screenshot:
 
 ```ruby
 browser = Ferrum::Browser.new
-browser.goto("https://google.com")
+browser.go_to("https://google.com")
 browser.screenshot(path: "google.png")
 browser.quit
 ```
@@ -91,7 +91,7 @@ Interact with a page:
 
 ```ruby
 browser = Ferrum::Browser.new
-browser.goto("https://google.com")
+browser.go_to("https://google.com")
 input = browser.at_xpath("//input[@name='q']")
 input.focus.type("Ruby headless driver for Chrome", :Enter)
 browser.at_css("a > h3").text # => "rubycdp/ferrum: Ruby Chrome/Chromium driver - GitHub"
@@ -102,7 +102,7 @@ Evaluate some JavaScript and get full width/height:
 
 ```ruby
 browser = Ferrum::Browser.new
-browser.goto("https://www.google.com/search?q=Ruby+headless+driver+for+Capybara")
+browser.go_to("https://www.google.com/search?q=Ruby+headless+driver+for+Capybara")
 width, height = browser.evaluate <<~JS
   [document.documentElement.offsetWidth,
    document.documentElement.offsetHeight]
@@ -116,7 +116,7 @@ Do any mouse movements you like:
 ```ruby
 # Trace a 100x100 square
 browser = Ferrum::Browser.new
-browser.goto("https://google.com")
+browser.go_to("https://google.com")
 browser.mouse
   .move(x: 0, y: 0)
   .down
@@ -194,7 +194,7 @@ Navigate page to.
   configuring driver.
 
 ```ruby
-browser.goto("https://github.com/")
+browser.go_to("https://github.com/")
 ```
 
 #### back
@@ -202,7 +202,7 @@ browser.goto("https://github.com/")
 Navigate to the previous page in history.
 
 ```ruby
-browser.goto("https://github.com/")
+browser.go_to("https://github.com/")
 browser.at_xpath("//a").click
 browser.back
 ```
@@ -212,7 +212,7 @@ browser.back
 Navigate to the next page in history.
 
 ```ruby
-browser.goto("https://github.com/")
+browser.go_to("https://github.com/")
 browser.at_xpath("//a").click
 browser.back
 browser.forward
@@ -223,7 +223,7 @@ browser.forward
 Reload current page.
 
 ```ruby
-browser.goto("https://github.com/")
+browser.go_to("https://github.com/")
 browser.refresh
 ```
 
@@ -232,7 +232,7 @@ browser.refresh
 Stop all navigations and loading pending resources on the page
 
 ```ruby
-browser.goto("https://github.com/")
+browser.go_to("https://github.com/")
 browser.stop
 ```
 
@@ -249,7 +249,7 @@ provided node.
     * :within `Node` | `nil`
 
 ```ruby
-browser.goto("https://github.com/")
+browser.go_to("https://github.com/")
 browser.at_css("a[aria-label='Issues you created']") # => Node
 ```
 
@@ -264,7 +264,7 @@ document or provided node.
   * :within `Node` | `nil`
 
 ```ruby
-browser.goto("https://github.com/")
+browser.go_to("https://github.com/")
 browser.css("a[aria-label='Issues you created']") # => [Node]
 ```
 
@@ -277,7 +277,7 @@ Find node by xpath.
   * :within `Node` | `nil`
 
 ```ruby
-browser.goto("https://github.com/")
+browser.go_to("https://github.com/")
 browser.at_xpath("//a[@aria-label='Issues you created']") # => Node
 ```
 
@@ -290,7 +290,7 @@ Find nodes by xpath.
   * :within `Node` | `nil`
 
 ```ruby
-browser.goto("https://github.com/")
+browser.go_to("https://github.com/")
 browser.xpath("//a[@aria-label='Issues you created']") # => [Node]
 ```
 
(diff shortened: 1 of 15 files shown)


AlexWayfer (changes requested):
Please, reread the #106: 

> For backward compatibility, I suggest not to rename, but add the alias.

So, I can't approve your PR with renaming.

alexanderadam:
Sure, but may I ask what's the reason for keeping `goto`?
IMHO it's not very rubyish and even in the ruby core api a lot of aliases got removed for consistency.
Personally I don't see a huge advantage in keeping `goto`.

AlexWayfer:
> Sure, but may I ask what's the reason for keeping `goto`?

I've mentioned the reason already twitce:

> > For backward compatibility

---

> IMHO it's not very rubyish

I agree, we all know that, but there are exclusions for third-party API.

For example: https://ruby-doc.org/core-2.7.2/Time.html#method-i-strftime

`strftime` is more C method (or even UNIX), so we have this instead of something like `format` or `format_time` or `format_time_as_string`.

Or: https://ruby-doc.org/core-2.7.2/String.html#method-i-casecmp-3F

It's not like `case_compared?`, but… you see.

So, `goto` is native name for Chromium API, you can see the same for `puppeteer`: https://github.com/puppeteer/puppeteer/blob/v5.4.1/docs/api.md#pagegotourl-options

In JavaScript, it should be `goTo` with considering `camelCase`, but… it's `goto` too.

---

> and even in the ruby core api a lot of aliases got removed for consistency.

Some examples? Anyway, I've described the reason with third-party API. Also, I think it's OK to have such alias instead of breaking existing projects using this gem.

---

> I mean backward compatibility is still given — it's still working but showing a deprecation notice suggesting to move to go_to (see the comment above yours for details).

Oh, sorry, didn't understand that. Hm… 

OK, give us some examples about "in the ruby core api a lot of aliases got removed for consistency", please, than I think we can consider such deprecation. Until this, I think my arguments about third-party API make sense.

ttilberg:
I just wanted to add my personal feedback that having both methods would be preferable to removing `goto` because of the compatibility with Puppeteer. While this isn't a Puppeteer clone, the parity is nice to have. It also matches [Watir](http://watir.com)'s method. As a Rubyist consuming the method, I would still think of `goto` before `go_to`...  perhaps from learning BASIC 30 years ago??? 😆

All this is to say that I support keeping this backwards compatible for a variety of reasons. Existing examples (however few) on the internet don't break, and `goto` is relevant in related libraries.

alexanderadam:
Okay, I removed [the deprecation warning](https://github.com/rubycdp/ferrum/pull/120/files#diff-2f9537c613f48bed5ff6ca273e2dac28886a32fe78dcd2dc7c7c798c891434d3R86) just now. At least the documentation in the README is also pointing to the variant with the Ruby convention now…

> I've mentioned the reason already twitce

And I also mentioned twice that backward compatibility is still given (my first two comments), since it's just showing a deprecation notice for now. :wink: 
I even showed a screenshot of the deprecation warning in the first comment.

Regarding
> `goto` is native name for Chromium API, you can see the same for `puppeteer`: [https://github.com/puppeteer/puppeteer/blob/v5.4.1/docs/api.md#pagegotourl-options](https://github.com/puppeteer/puppeteer/blob/v5.4.1/docs/api.md#pagegotourl-options)

and
> having both methods would be preferable to removing `goto` because of the compatibility with Puppeteer. While this isn't a Puppeteer clone, the parity is nice to have.

Chrome / Puppeteer are using JavaScript based APIs. The JavaScript naming convention is obviously _not_ the same as the Ruby convention. And even for JavaScript itself it's inconsistent because it's usually using camecased function names (`newPage`, `createBrowserFetcher`, `createIncognitoBrowserContext`, `waitForTarget`, …). IMHO _all_ multi word functions are in camel case _except_ `goto`.
We are obviously not the first people wondering about this, since this is also [an unanswered StackOverflow question from two years ago](https://stackoverflow.com/questions/53089928/puppeteer-goto-method-not-in-camelcase).

Independent from following a _bad_ example is still _leaving it_ as a bad example, isn't it? It sounds a bit like "_Oh, someone did it in an inconsistent way, therefore we should also do it inconsistently_". It really doesn't help and won't improve something if we're following blindly. Ruby has pretty good conventions that are used and known in the community. What exactly is the advantage on knowingly _breaking_ a convention for a _single_ method and just bringing the excuse that someone else _also_ did it wrong?

Apart from that: if Puppeteer API compatibility (or any other inconsistent API) should be one of the goals of Ferrum, it should change the case of other methods to camel case as well. Otherwise it is some kind of double inconsistency, isn't it? :wink: 
Or is this a useful credo: "_We are following the naming convention of another project — but only for a **single** method._"?

Furthermore I think we should leave every place _better_ than we found it. Including source code. Thus: if I find some inconsistencies, I would try to fix it, instead of keeping it (and making other people wondering about it).

> I support keeping this backwards compatible for a variety of reasons. Existing examples (however few) on the internet don't break, and `goto` is relevant in related libraries.

Again, as mentioned before:

>  backward compatibility is still given — it's still working but showing a deprecation notice suggesting to move to `go_to`

It follows a common deprecation pattern. It wasn't going to break — it would have shown a deprecation warning.
So if you are a developer who's updating dependencies every now and then, you would see the warning and can safely migrate to the method name that follows the Ruby convention. Without. breaking. anything.

Chances are high, that you saw deprecation warnings in other projects and methods anyway. This is a pretty common thing.

route:
Guys calm down, I'm fine with both namings. So let's just add an alias to `go_to` and everyone will be happy. Wanna use `goto` fine, `go_to` excellent. Up to end user.

alexanderadam:
I'm sure we're all calm. I hope no comment here sounded harsh‽ Please excuse me if any comment did!

The current PR does not deprecate anything any more, however.
Both names are available but the method following the Ruby convention is used internally and in the documentation.
The warning [mentioned in the first comment](https://github.com/rubycdp/ferrum/pull/120#issue-510146489) is not present anymore.

route:
Looks good to me, thank you!

AlexWayfer:
> And I also mentioned twice that backward compatibility is still given (my first two comments), since it's just showing a deprecation notice for now. :wink: 

OK, it's a point.

> Chrome and Puppeteer are using JavaScript based APIs.

It's controversial: Chrome's `goto` is not JavaScript API, it's C/C++ code, as I understand.

> The JavaScript naming convention is obviously not the same as the Ruby convention.

Yes: it's `camelCase` instead of `snake_case`, but `goto` is out of both of them, it's something like keyword I guess from old programming languages.



> It sounds a bit like "_Oh, someone did it in an inconsistent way, therefore we should also do it inconsistently_".

I've created #106 for something like "Hey, let's do this thing from another ecosystem in *our consistent way*", so no.



> Ruby has pretty good conventions that are used and known in the community.

And JavaScript too: `camelCase`. But again: `goto` is a kind of exception. And I'm not against your PR, I just want to save backward compatibility: between our (Ruby) consistent and their (who has intended and using `goto`). Again: I didn't understand your deprecation point, that this form of method is still available, but we should discuss it before merge — this is my point. Like… you've created a PR with auto-resolving an issue with different behavior. I'm against this: let's discuss should we support both method names or should we deprecate one of them and then create a PR or change an issue. That's all.



> Apart from that: if Puppeteer API compatibility (or any other inconsistent API) should be one of the goals of Ferrum, it should change the case of other methods to camel case as well. Otherwise it is some kind of double inconsistency, isn't it?

Also a good point, but again: `goto` seems as exception for any language, ecosystem, etc.: it's not `camelCase` neither `snake_case`. And it's already here, in Ferrum. So, we should discuss it. Again: I just tried to point the difference between the linked issue and this PR, no more.



> Or is this a useful credo: "_We are following the naming convention of another project — but only for a **single** method._"?

Maybe, because `goto` is a keyword for many languages like `for`, `if`, end`, etc. It's a more difficult case, I understand, but… I have no more ideas why it was called so.



> Guys calm down, I'm fine with both namings.

I'm calm, don't worry. :sweat_smile: I hope you're all got my refined point.



> I hope no comment here sounded harsh‽

For me every your comment is OK, I hope my comments are also OK for you and others. I just like to discuss and understand each other. 😬

Issue #135, opened 2021-01-08
Handling of canceled requests


julioolvr:
I'm still not 100% sure about this, but I've been trying to ensure all requests where handled before wrapping up each test case with `page.driver.wait_for_network_idle` in an `after` hook. Even though using Chrome in non-headless mode I was able to see al requests being completed, according to Ferrum the network still wasn't idle.

I grabbed the pending requests with `page.driver.network_traffic.select(&:pending?)` and started comparing and noticed that the supposedly pending requests where the ones cancelled by the browser (for instance, by using `AbortController` to cancel a request done with `fetch`).

I found the [`loadingFailed`](https://chromedevtools.github.io/devtools-protocol/tot/Network/#event-loadingFailed) event in the devtools protocol so I tried something like this as a quick test:

```ruby
@page.on("Network.loadingFailed") do |params|
  exchange = select(params["requestId"]).last

  if exchange && params['canceled']
    puts 'Found exchange, setting arbitrary error'
    exchange.error = :canceled
  end
end
```

And that seems to properly consider canceled requests as "finished". There might be _a lot_ I'm overlooking since this is in the context of a big app with plenty of moving pieces but since I saw a clear improvement I figured I'd brought it up.

route:
That sounds right. I think I had a fight with requests before, if you could come up with a quick tests to run and see the issue or PR I'd gladly merge it, thanks for this investigation!

julioolvr:
Awesome, I'll give it a shot over the weekend 👍

route:
Closing this then!

PR #136, opened 2021-01-16
Canceled request handling

julioolvr:
This is a proof of concept for fixing #135. The spec fails initially and passes with the second commit.

How do you want to handle this? My initial suggestion would be for a canceled request to be considered as a failed network exchange (CDP's event is called `Network.loadingFailed` after all) and so I could assign a `Network::Error` instance to `exchange.error` but I haven't put much thought into this so I wanted to check first.

All that being said, I also see that errors are currently handled by `Log.entryAdded` instead of `Network.loadingFailed` so maybe there's something I'm missing. Otherwise I wonder if we should handle _all_ `Network.loadingFailed` events (instead of just those where `params['canceled'] == true`).

### DIFF
--- a/lib/ferrum/network.rb
+++ b/lib/ferrum/network.rb
@@ -157,6 +157,14 @@ module Ferrum
         end
       end
 
+      @page.on("Network.loadingFailed") do |params|
+        exchange = select(params["requestId"]).last
+
+        if exchange && params['canceled']
+          exchange.error = :canceled
+        end
+      end
+
       @page.on("Log.entryAdded") do |params|
         entry = params["entry"] || {}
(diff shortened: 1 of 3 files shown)


route:
First of all @julioolvr thank you so much for finding this bug and hunting it down. Second yes canceled request should be considered as failed. I think I added `Log.entryAdded` because it provides more info about failed request and I just didn't know that it's possible to have either `Network.loadingFailed` or both `Log.entryAdded` + `Network.loadingFailed`. I think it looks fine as it is now, I need to run test suit with this change for `cuprite`. I'll add `Network::Error` instead of a symbol today but I think we can merge it

PR #139, opened 2021-01-20
Fix test: resize viewport by fullscreen

Mifrill:
We have 2 things here:
1) The test `allows the viewport to be resized by fullscreen` with specified explicitly `width/height` is green for both cases: offsetWidth and scrollWidth
HTML:
```
<html>
  <head>
    <title>custom_size</title>
    <style>
      html {
        width: 1280px;
        height: 1024px;
      }
    </style>
  </head>
  <body>
  </body>
</html>
```
Green:
```
      def document_size
        evaluate <<~JS
          [document.documentElement.offsetWidth,
           document.documentElement.offsetHeight]
        JS
      end
```
Green:
```
      def document_size
        evaluate <<~JS
          [document.documentElement.scrollWidth,
           document.documentElement.scrollHeight]
        JS
      end
```

so, I think, we should keep test with this HTML.

2) The second case was described here: https://github.com/rubycdp/ferrum/pull/51

I think, it's better to use the dedicated file with HTML for `width: 100%; / height: 100%;`:

So, test failed:
```
expected: [1280, 1024]
            got: [1280, 1016]
```

because of default Chrome css margins:
https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/core/html/resources/html.css;l=57-60

Which was added, I guess, in `72.0.3626.0` version:
https://source.chromium.org/chromium/chromium/src/+/refs/tags/72.0.3626.0:third_party/blink/renderer/core/html/resources/html.css
Source

https://source.chromium.org/chromium/chromium/src/+/refs/tags/72.0.3625.2:third_party/blink/renderer/core/html/resources/html.css
No source for the previous version.

We also, can confirm it here: https://browserdefaultstyles.com/
Selection_046

So, the correct values for this test, should be:
Selection_045

### DIFF
--- a/spec/browser_spec.rb
+++ b/spec/browser_spec.rb
@@ -104,7 +104,7 @@ module Ferrum
 
       include_examples 'resize viewport by fullscreen' do
         let(:path) { "/ferrum/custom_html_size_100%" }
-        let(:viewport_size) { [1280, 1016] }
+        let(:viewport_size) { [1272, 1008] }
       end
     end
 
--- a/spec/browser_spec.rb
+++ b/spec/browser_spec.rb
@@ -87,11 +87,25 @@ module Ferrum
       expect(browser.viewport_size).to eq([200, 400])
     end
 
-    it "allows the viewport to be resized by fullscreen" do
-      browser.go_to("/ferrum/custom_html_size")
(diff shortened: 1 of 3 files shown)


route:
Oh awesome! Nice job! Thank you!

PR #141, opened 2021-01-21
restore specs with 'element is not in the viewport' context

Mifrill:
- Implement `Ferrum::Node#school_into_view`
- Implement `Ferrum::Node#in_viewport?`
- restore specs with context: `when the element is not in the viewport of parent element`


lib/ferrum/node.rb
```diff
@@ -46,6 +46,8 @@ def click(mode: :left, keys: [], offset: {}, delay: 0)
       modifiers = page.keyboard.modifiers(keys)
 
       case mode
+      when :force
+        evaluate("this.click()")
```
Mifrill, line 50: it's based on real clicks: 

> Puppeteer also provides two methods to deliver real clicks to the elements:page.click('.foo') and elementHandle.click(). These, however, have a little bit of trickery underneath.

https://medium.com/@aslushnikov/automating-clicks-in-chromium-a50e7f01d3fb

and on the `force: true` option in Cypress, for example:
```
cy.get('input[type="search"]').click({force: true})
```

spec/node_spec.rb
```diff
@@ -65,20 +65,19 @@ module Ferrum
...
 
-    context "when the element is not in the viewport of parent element", skip: true do
+    context "when the element is not in the viewport of parent element" do
       before do
         browser.go_to("/ferrum/scroll")
       end
 
       it "scrolls into view" do
-        # FIXME:
-        browser.at_xpath("//a[text() = 'Link outside viewport']").click
```
Mifrill, line 75: @route is the idea was: use `click` without options even if the element is outside of viewport?

like this:
```
if element_outside_of_viewport?
  evaluate("this.click()")
  return self
end

def element_outside_of_viewport?
  evaluate("this.getBoundingClientRect().bottom < 0") ||
    evaluate("this.getBoundingClientRect().right < 0") ||
    evaluate("this.getBoundingClientRect().left > (window.innerWidth || document.documentElement.clientWidth)") ||
    evaluate("this.getBoundingClientRect().top > (window.innerHeight || document.documentElement.clientHeight)")
end
```

based on https://stackoverflow.com/questions/123999/how-can-i-tell-if-a-dom-element-is-visible-in-the-current-viewport/7557433#7557433

spec/node_spec.rb
```diff
@@ -65,20 +65,19 @@ module Ferrum
...
-        expect(browser.current_url).to eq("/")
+        browser.at_xpath("//a[text() = 'Link outside viewport']").click(mode: :force)
+        expect(browser.current_url).to eq(base_url("/"))
       end
 
       it "scrolls into view if scrollIntoViewIfNeeded fails" do
```
Mifrill, line 78: @route Those tests have references into `scrolls` is it based on?

> Indeed, we can't click things that are not on screen, so we try to scroll them in. If website has a custom logic around scrolling, this might break.

https://github.com/puppeteer/puppeteer/issues/2190#issuecomment-380254352

we probably may change tests descriptions due to this approach:

> I think it might be worth noting in the documentation that all of the click/mouse functionality depends on emulating real clicks where the link needs to be visible.

https://github.com/puppeteer/puppeteer/issues/2190#issuecomment-380262695

lib/ferrum/node.rb
route: This one is synthetic event and has nothing in common with real clicks

spec/node_spec.rb
route: Check `cuprite` on how it's handled there. I think main difference is that `cuprite` scrolls into viewport automatically because it's a complicated driver, `ferrum` instead should be close to what real user would do with the browser. So if you can't see a node in the viewport you can't click on it, but we might be missing some API for convenient scrolling then.. I think we can drop these tests in general after having something for scrolling but I left them as a reminder to implement it.

spec/node_spec.rb
route: Didn't get what you mean

Mifrill: I was mean: the descriptions of the skipped tests are included scroll-references, like: `scrolls into view ...`, so, as I understand, the fix of those tests should contain some of `real clicks` implementation or we can use the emulating with avoid the scroll and, probably, with the change of the descriptions.

This is the main sentence which I was needed to choose the right way:

> This one is synthetic event and has nothing in common with real clicks

so, `synthetic event` about
> emulating real clicks where the link needs to be visible.

real clicks about:
scroll into element (somehow) to click on it

lib/ferrum/node.rb
```diff
@@ -88,6 +88,26 @@ def hover
       raise NotImplementedError
     end
 
+    def scroll_into_view
+      tap { page.command("DOM.scrollIntoViewIfNeeded", nodeId: node_id) }
```
Nakilon, line 92: I use it like this:
```ruby
  scroll_and_click = ->_{ Browser.page.command "DOM.scrollIntoViewIfNeeded", nodeId: _.node_id; _.click }
  scroll_and_click.call Browser.at_css(".mdc-select")
```
and sometimes it throws:
> /Users/nakilon/.rbenv/versions/2.5.9/lib/ruby/gems/2.5.0/gems/ferrum-0.11/lib/ferrum/browser/client.rb:96:in `raise_browser_error': Node does not have a layout object (Ferrum::BrowserError)

route:
Manually merged

PR #143, opened 2021-01-27
Implement focusable?

### LINKED ISSUE #90: Set value for non-focusable input
Is there a way how to fill hidden input field? Eg. a TEXTAREA with `style="display: none;"`.

When I do this:

```
page.at_css('#hidden-field').focus.type('foobar')
```

I get `Ferrum::BrowserError: Element is not focusable`

Currenly this is working for me, but some native way of setting hidden values would be nice.

```
page.execute("document.getElementById('hidden-field').value = 'foobar'")
```

Is there a better way?

Mifrill:
closes #90

### DIFF
--- a/lib/ferrum/node.rb
+++ b/lib/ferrum/node.rb
@@ -30,6 +30,13 @@ module Ferrum
       tap { page.command("DOM.focus", slowmoable: true, nodeId: node_id) }
     end
 
+    def focusable?
+      focus
+      true
+    rescue BrowserError => e
+      e.message == "Element is not focusable" ? false : raise
+    end
+
     def blur
       tap { evaluate("this.blur()") }
     end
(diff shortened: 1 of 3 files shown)


route:
I have a few doubts about this PR, I'll write an explanation later after confirming.

route (commented):
I think it can be problematic to get and name all those methods that we can extract from styles... This style is only one from a mix that determines if element is visible and focusable. Check Cuprite for `isVisible`. I think we could provide some instruments to check visibility and focusability but it all depends on Chrome API to do that properly. In this case I'd add `focusable?` and with rescue checked and return boolean.

lib/ferrum/node.rb
```diff
@@ -62,6 +62,18 @@ def click(mode: :left, keys: [], offset: {}, delay: 0)
       self
     end
 
+    def set_attribute(value, name:)
```
route, line 65: Before adding such a method I wanted first to make Ferrum/Cuprite API stable for this kind of methods, there's now a confusion with attributes, properties and `[](name)` see in Cuprite node.rb

PR #144, opened 2021-01-27
Specify slowmo as delay in seconds

ttilberg:
I went digging into the source to find if this was in seconds or ms since Puppeteer is always in ms. I felt it should just be added to the readme saving people the future dig.


Mifrill:
@ttilberg Thank you!

PR #148, opened 2021-02-02
Screenshot with specific background color

Mifrill:
closes #119

### DIFF
--- a/README.md
+++ b/README.md
@@ -335,7 +335,7 @@ Saves screenshot on a disk or returns it as base64.
   * :full `Boolean` whether you need full page screenshot or a viewport
   * :selector `String` css selector for given element
   * :scale `Float` zoom in/out
-  * :background_rgba_color `Array` like `[R,G,B,A]` to have specific background color
+  * :background_color `Ferrum::RGBA.new(0, 0, 0, 0.0)` to have specific background color
 
 ```ruby
 browser.go_to("https://google.com/")
--- a/README.md
+++ b/README.md
@@ -346,7 +346,7 @@ browser.screenshot(path: "google.jpg") # => 30902
 # Save to Base64 the whole page not only viewport and reduce quality
 browser.screenshot(full: true, quality: 60) # "iVBORw0KGgoAAAANSUhEUgAABAAAAAMACAYAAAC6uhUNAAAAAXNSR0IArs4c6Q...
 # Save with specific background color
-browser.screenshot(background_rgba_color: [0, 0, 0, 0.0]) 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -335,6 +335,7 @@ Saves screenshot on a disk or returns it as base64.
   * :full `Boolean` whether you need full page screenshot or a viewport
   * :selector `String` css selector for given element
   * :scale `Float` zoom in/out
+  * :background_rgba_color `Array` like `[R,G,B,A]` to have specific background color
```
route, line 338: Let's rename it to `background_color`

lib/ferrum/page/screenshot.rb
```diff
@@ -123,9 +123,21 @@ def screenshot_options(path = nil, format: nil, scale: 1.0, **opts)
...
         end
 
+        if validate_background_rgba_color(opts[:background_rgba_color])
+          raise ArgumentError, "Specify :background_rgba_color as [R,G,B,A] array"
+        end
+
         options
       end
 
+      def validate_background_rgba_color(option)
```
route, line 133: I don't feel like this code belongs to screenshot. Anonymous array is definitely a dynamic typing approach but maybe if we introduce something like

```ruby
browser.screenshot(background_rgba_color: [0, 0, 0, 0.0])

browser.screenshot(background_color: RGBA.new(0, 0, 0, 0.0))
```

we could hide validations in there as well as make this variable sound right

lib/ferrum/page/screenshot.rb
```diff
@@ -159,6 +173,17 @@ def maybe_resize_fullscreen(full)
       ensure
         resize(width: width, height: height) if full
       end
+
+      def with_background_color(background_rgba_color)
+        if background_rgba_color
+          r, g, b, a = background_rgba_color
+          command('Emulation.setDefaultBackgroundColorOverride', color: { r: r, g: g, b: b, a: a })
```
route, line 180: Then we could make it `command('Emulation.setDefaultBackgroundColorOverride', color: color.to_h)` if it's an instance of RGBA

spec/screenshot_spec.rb
```diff
@@ -176,6 +176,46 @@ def create_screenshot(**options)
           end
         end
 
+        context 'background_rgba_color option' do
+          shared_examples 'background_rgba_color ArgumentError' do
+            it 'raises ArgumentError with message' do
```
route, line 181: and we could clean up tests as we only accept RGBA or raise

route (approved):

PR #150, opened 2021-02-03
issue_149: redefine existed 'on'-handler in Browser::Subscriber

### LINKED ISSUE #149: Resources still retrieved when expressly rejected using callbacks when using authenticated proxies
Loving the gem so far. Much easier to deal with Ruby.

I ran into this problem when I was attempting to troubleshoot other things. 

I'm using callbacks to limit what is retrieved to limit bandwidth consumption (images, etc). I'm also using proxies.

In testing, I realized that my screenshots still had images that are expressly rejected in the callback. I added some code to show that the DENY was still firing in the callback.  

But, the images were still there. 

So, I wrote a separate script that can be used to test this case (below). 

If you don't use a proxy, everything works as expected and images are not present in the screenshot. 

If you add an authenticating proxy, the images and other callbacks show that they fire and are rejected. But, the images are still present. 

It's possible that I'm misconfiguring this. If so, please help me sort out what I'm doing wrong. But, if not, then it's a bug. 

In the code below, I show this. You can comment out the proxy details and re-run the script. Without a proxy, it will reject the images. If you put a proxy in, the images will still be present in the screenshot. 

If I've messed up my config and it's a misconfiguration problem, please let me know! 

```
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'ferrum'
end

require 'ferrum'

allowed_types = %w[Document]
headers = {"User-Agent"=>"Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/87.0.4280.88 Safari/537.36"}

# Remove the proxy server option from the hash to test without the proxy server
retriever_options = {:browser_options=>{"proxy-server"=>"XX:XX:XX:XX:39393", "ignore-certificate-errors"=>nil}, :timeout=>30, :xvfb=>false}

browser = ::Ferrum::Browser.new(retriever_options)
browser.headers.set(headers)

# Remove this line when you are testing without the proxy.
browser.network.authorize(type: :proxy, user: 'XXXXXXX', password: 'XXXXXXX')

browser.network.intercept
browser.on(:request) do |request|
  if allowed_types.include?(request.resource_type)
    puts "ALLOW #{request.resource_type} #{request.url}"
    request.continue
  else
    puts "DENY #{request.resource_type} #{request.url}"
    request.abort
  end
end

# Change this to the site you want to test with.
url = "https://site.com/"

browser.goto(url)
sleep 4.0 # I put this in just to make sure all JS is done rendering, doesn't affect the case. 
browser.screenshot(path: 'site.html.jpg')
```

Mifrill:
closes #149

### DIFF
--- a/lib/ferrum/browser/subscriber.rb
+++ b/lib/ferrum/browser/subscriber.rb
@@ -17,11 +17,7 @@ module Ferrum
       end
 
       def on(event, &block)
-        if @on[event].empty?
-          @on[event] << block
-        else
-          @on[event] = [block]
-        end
+        @on[event] = [block]
         true
       end
 
--- a/lib/ferrum/browser/subscriber.rb
+++ b/lib/ferrum/browser/subscriber.rb
@@ -17,7 +17,11 @@ module Ferrum
       end
 
       def on(event, &block)
-        @on[event] << block
+        if @on[event].empty?
+          @on[event] << block
+        else
+          @on[event] = [block]
+        end
         true
       end
 
--- a/spec/network_spec.rb
+++ b/spec/network_spec.rb
@@ -328,6 +328,18 @@ module Ferrum
         expect(browser.network.status).to eq(200)
         expect(browser.body).to include("content")
       end
+
+      it "redefines existed handler" do
+        browser.network.intercept
+        browser.on(:request) do |request|
+          request.abort
+        end
+        browser.on(:request) do |request|
+          request.continue
+        end
+        browser.go_to("/ferrum/url_blacklist")
+        expect(browser.network.status).to eq(200)
+      end
     end
   end
 end


lib/ferrum/browser/subscriber.rb
```diff
@@ -17,7 +17,7 @@ def initialize
       end
 
       def on(event, &block)
-        @on[event] << block
+        @on[event] = [block]
```
Mifrill, line 20: @route we have 2 subscribers for `Network.requestWillBeSent`, do we have missed test for this?

route: Yes we should have an ability to define a few handlers.

* `Network.requestWillBeSent` in `frames.rb` is defined because we want to know if Chrome is going to make request for the main frame for the whole document (the whole page), if it does then we reset `@event` because now we shouldn't return back to the main Thread until page is loaded which signaled by Chrome.
* `Network.requestWillBeSent` in `network.rb` is used to define request and exchange instances that are used by people listening to network traffic

So in other words if you leave only one you won't be able to build correctly either network traffic or you will return immediately even though page is only started to load or reload.

route: I don't see that it's wrong in any sense to have an ability to have a few handlers. I think that current implementation of authentication needs reconsidering, though it will be hard because we only use what Chrome provides.

route: A lot of stuff is missing in current version of PR, it's not ready in any sense. I'm closing so that people don't use it because it breaks the gem, while tests are passing.

Mifrill: > we should have an ability to define a few handlers

well, okay, then we may, probably, use the `redefine:` flag for `on` method to allow both cases:
- define a few handlers
- redefine existed handler

Mifrill: so, with the flag usage, we can keep the current behavior of handlers is unchanged, and provide the fix for #149 in this way:
```
browser.on(:request, redefine: true) do |request|
  if allowed_types.include?(request.resource_type)
    puts "ALLOW #{request.resource_type} #{request.url}"
    request.continue
  else
    puts "DENY #{request.resource_type} #{request.url}"
    request.abort
  end
end
```
@route WDYT?

route: Yep let me sleep with it for tomorrow.

route: Decided to go with https://github.com/rubycdp/ferrum/pull/154, still not finished but I think that would work

Issue #132, opened 2020-12-02
Equivalent of Puppeteer's `evaluateOnNewDocument`


alexanderadam:
See the [`evaluateOnNewDocument`](https://github.com/puppeteer/puppeteer/blob/b57f3fcd5393c68f51d82e670b004f5b116dcbc3/docs/api.md#pageevaluateonnewdocumentpagefunction-args) documentation for details.

Interesting bit for me is, that

> [T]he function is invoked after the document was created but before any of its scripts were run.

This makes it pretty mighty and it seems that [other Ruby libraries aren't supporting this at the moment yet](https://stackoverflow.com/questions/57487802/watir-equivalent-to-evaluateonnewdocument).

route:
This is what `extensions` are for https://github.com/rubycdp/ferrum#customization

alexanderadam:
Any particular reason what would speak against a native integrated `evalulate_on_new_document` or similar?

route:
None, I started Cuprite project first and tried to make it compatible with Poltergeist, thus :extensions appeared. This project was extracted from Cuprite, I don't mind having dedicated method, if you open a PR gladly merge it.

alexanderadam:
Okay, good. Would you consider reopen this issue then please? :wink:

skyeagle:
just in a case, someone is looking for an example of doing it with [cuprite](github.com/rubycdp/cuprite) and extensions:

```ruby
Capybara.register_driver(:cuprite) do |app|
  Capybara::Cuprite::Driver.new(
    app,
    window_size: [1200, 800],
    extensions: [
      { source: "console.log('Page.addScriptToEvaluateOnNewDocument triggered')" },
    ],
  )
end
```

PR #161, opened 2021-02-19
fixes issue #132 add evalulate_on_new_document

alexanderadam:
Hi @route,

so I thought that I might give [issue 132](https://github.com/rubycdp/ferrum/issues/132) a go.

As always: thank you for your work and [never give up](https://twitter.com/alexanderadam__/status/1362818462557622272) :wink:

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,7 +124,6 @@ browser.mouse
 browser.quit
 ```
 
-
 ## Docker
 
 In docker as root you must pass the no-sandbox browser option:
@@ -775,6 +774,18 @@ browser.add_script_tag(content: "window.__injected = 42")
 browser.evaluate("window.__injected") # => 42
 ```
 
+#### evaluate_on_new_document(javascript)
+
+Evaluate JavaScript to modify things before a page load:
+
+```ruby
+javascript = <<~JS
+    Object.defineProperty(navigator, 'languages', {
+      get: function() { return ['tlh']; }
+    });
+  JS
+browser.evaluate_on_new_document javascript
+```
 
 ## Frames
 
--- a/lib/ferrum/browser.rb
+++ b/lib/ferrum/browser.rb
@@ -77,6 +77,10 @@ module Ferrum
       end
     end
 
+    def evaluate_on_new_document(javascript)
+      extensions << javascript
+    end
+
     def timeout
       @timeout || DEFAULT_TIMEOUT
     end
--- a/spec/browser_spec.rb
+++ b/spec/browser_spec.rb
@@ -205,6 +205,25 @@ module Ferrum
           browser&.quit
         end
       end
+
+      it 'supports user Javascript' do
+        begin
+          javascript = <<~JS
+            Object.defineProperty(navigator, 'languages', {
+              get: function() { return ['tlh']; }
+            });
+          JS
+
+          browser = Browser.new.tap { |b| b.evaluate_on_new_document(javascript) }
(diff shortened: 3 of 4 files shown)


README.md
```diff
@@ -124,6 +124,18 @@ browser.mouse
...
+Evaluate JavaScript to modify things before a page load:
+
+```ruby
+javascript = <<~JS
+  Object.defineProperty(navigator, 'languages', {
+    get: function() { return ['tlh']; }
+  });
+JS
+
+Browser.new(evaluate_on_new_document: javascript)
```
route, line 136: I think we already have `:extensions` we can say in the description of this newly added method that they are related, no need to introduce new option.

lib/ferrum/browser.rb
```diff
@@ -77,6 +77,10 @@ def extensions
       end
     end
 
+    def evaluate_on_new_document
```
route, line 80: I think we should reuse what we have for extensions now and modify `@extensions` array calling this method

README.md
alexanderadam: I'm not quite sure how you meant this exactly but I changed it how I understood it.
Can you check again please?

lib/ferrum/browser.rb
```diff
@@ -77,6 +77,11 @@ def extensions
       end
     end
 
+    def evaluate_on_new_document(javascript = nil)
+      extensions << (block_given? ? yield : javascript)
```
route, line 81: block is not needed I guess, and we definitely should accept something so not need for `= nil`

lib/ferrum/browser.rb
```diff
@@ -77,6 +77,11 @@ def extensions
...
+    end
+    alias evaluate_on_new_document= evaluate_on_new_document
```
route, line 83: `evaluate_on_new_document=` confuses me as I would think that I can override previously passed script, let's leave only `evaluate_on_new_document`

README.md
```diff
@@ -124,6 +124,18 @@ browser.mouse
 browser.quit
 ```
 
+Evaluate JavaScript to modify things before a page load:
```
route, line 127: We have a section for JavaScript, let's move this method there

lib/ferrum/browser.rb
alexanderadam: Oh really? I liked that syntax:

```ruby
browser.evaluate_on_new_document do
  <<~JS
    Object.defineProperty(navigator, 'languages', {
      get: function() { return ['tlh']; }
    });
  JS
end
```

I changed it accordingly though.

ttilberg: That syntax is almost still entirely valid, however by adding the block into the mix, there's a certain element of unnecessary indirection in the API. Questions that come to mind are "Does the block evaluate in a certain context? Are there args? How is this different from the `javascript` arg?"

You surely know this, but this is still valid:

```rb
browser.evaluate_on_new_document <<~JS
  Object.defineProperty(navigator, 'languages', {
    get: function() { return ['tlh']; }
  });
JS
```

And I leaves fewer questions when learning the API.

As a side note, I too love using the heredoc+highlight syntax for `js` strings. Particularly in Ferrum, it feels fantastic.

alexanderadam: > by adding the block into the mix, there's a certain element of unnecessary indirection in the API

For me blocks create a semantic boundary which goes further than "only" having a heredoc.
But again: I changed it anyway. :wink:

route:
Looking good, thank you!

PR #184, opened 2021-06-23
document Browser#reset

igneus:
It took me quite a lot of time to find out how to use a long-running Chrome process without all browser tabs ever created hanging in Chrome forever. I think `Browser#reset` deserves a mention in the documentation.


route:
Thanks

PR #189, opened 2021-08-20
Implement Tracing

### LINKED ISSUE #171: Implement Tracing
https://github.com/puppeteer/puppeteer/blob/v8.0.0/docs/api.md#class-tracing

Mifrill:
closes #171


lib/ferrum/browser/tracing.rb
```diff
@@ -0,0 +1,63 @@
...
+      def start(path: '', screenshots: false, categories: DEFAULT_CATEGORIES)
+        raise "Cannot start recording trace while already recording trace." if recording
+        categories = categories.concat(["disabled-by-default-devtools.screenshot"]) if screenshots
+        self.path = path
+        self.recording = true
+        @client.command("Tracing.start", transferMode: "ReturnAsStream", categories: categories.join(','))
+      end
+
+      def stop
+        @client.on("Tracing.tracingComplete", handler_type: :once) do |event|
```
Mifrill, line 35: @route this event is not triggered for some of the reasons, could you put the advice, please?

route: Because test finishes faster than the event comes, and you most likely need @client for page not for the whole browser.

lib/ferrum/browser/subscriber.rb
```diff
@@ -33,6 +33,24 @@ def call(message)
...
         end
       end
+
+      private
+
+      def handle_any_times(event, &block)
+        @on[event] << block
+      end
+
+      def handle_once(event, &block)
```
route, line 43: This is very unreadable and hard, if I were you I'd start refactoring it first before moving further. Not sure if I like the idea of having this `:handler_type` on event listener. Can you explain why we need it in the first place?

lib/ferrum/browser/tracing.rb
Mifrill: Oh wow :tada:, indeed, I didn’t think about it, I thought it’s a matter of configuration. @route Thank you! :muscle:

> page not for the whole browser, because each page should be traced independently

yeah, it makes sense, let's use `page`.

Just a note about `page`:
I see in playwright implementation for `page` reference, for example:
```js
this._tracingClient = page ? (page._delegate as CRPage)._mainFrameSession._client : this._session;
```
Source: https://github.com/microsoft/playwright/blob/c373986ca0b91b76d4910ce0617b55e50934114c/packages/playwright-core/src/server/chromium/crBrowser.ts#L240

however for `puppeteer`, they use the [connection](https://github.com/puppeteer/puppeteer/blob/040a0e561b4f623f7929130b90be129f94ebb642/src/common/Connection.ts) with no specific `page` reference:
```js
export class Tracing {
  _client: CDPSession;
```
Source: https://github.com/puppeteer/puppeteer/blob/040a0e561b4f623f7929130b90be129f94ebb642/src/common/Tracing.ts#L45

lib/ferrum/browser/subscriber.rb
Mifrill: > if I were you I'd start refactoring it first before moving further

yeah, of course, it's a Draft PR with [WIP] code, just the attempt to make it works.

> Can you explain why we need it in the first place?

sure, the original intention is to follow the existing implementation in [puppeteer link that you provided](https://github.com/rubycdp/ferrum/issues/171#issue-823832394).

Those guys using `once` type of handler just for a few cases, for example:
image
At the same time they use a handler for any_times for such more cases:
image

Hence, they have an intention to use a handler with a `once` type for `Tracing.tracingComplete` event.

Apart from that, the `microsoft/playwright` implementation of tracing also use a handler with `once` type:
```ts
new Promise(f => this._tracingClient!.once('Tracing.tracingComplete', f)),
```
Source: https://github.com/microsoft/playwright/blob/c373986ca0b91b76d4910ce0617b55e50934114c/packages/playwright-core/src/server/chromium/crBrowser.ts#L268

I'm not sure about the mandatory of use the once-type event handler for `Tracing.tracingComplete` event, but it looks reasonable to handle events only once by `stop` command.

lib/ferrum/browser/tracing.rb
Mifrill: About:
> each page should be traced independently

```js
  it('should throw if tracing on two pages', async () => {
    await page.tracing.start({ path: outputFile });
    const newPage = await browser.newPage();
    let error = null;
    await newPage.tracing
      .start({ path: outputFile })
      .catch((error_) => (error = error_));
    await newPage.close();
    expect(error).toBeTruthy();
    await page.tracing.stop();
  });
```
Source: https://github.com/puppeteer/puppeteer/blob/040a0e561b4f623f7929130b90be129f94ebb642/test/tracing.spec.ts#L93-L103

So, we probably should avoid using tracing for a few pages at the same time. @route What do you think?

route: Maybe, I don't know because I need to start playing around to prove it. I trust you

lib/ferrum/browser/subscriber.rb
route: I still don't understand the intention. If I subscribe to `Tracing.tracingComplete` does it mean it is executed multiple times? If so what are the implications of this?

Mifrill: > what are the implications of this

great question! well, as I understand, the issue is playable for the case when we use tracing a few times for the same page instance.

For example, we wanna trace events -> then stop tracing ->, then do whatever something ->, and then start new tracing but for a different file:
```ruby
context "handle tracing.Complete event once" do
  it "outputs a trace" do
    file_path = "#{PROJECT_ROOT}/spec/tmp/trace.json"
    browser.tracing.start(path: file_path)
    browser.go_to("https://www.google.com")
    browser.tracing.stop
    sleep 1
    expect(File.exist?(file_path)).to be(true)
	
    file_path2 = "#{PROJECT_ROOT}/spec/tmp/trace2.json"
    browser.tracing.start(path: file_path2)
    browser.go_to("https://www.facebook.com")
    browser.tracing.stop
    sleep 1
    expect(File.exist?(file_path2)).to be(true)

    file_path3 = "#{PROJECT_ROOT}/spec/tmp/trace3.json"
    browser.tracing.start(path: file_path3)
    browser.go_to("https://github.com")
    browser.tracing.stop
    sleep 1
    expect(File.exist?(file_path3)).to be(true)
  ensure
    FileUtils.rm_f(file_path)
    FileUtils.rm_f(file_path2)
    FileUtils.rm_f(file_path3)
  end
end
```

here is a difference:
```ruby
def stop
  client.on("Tracing.tracingComplete") do |event|
    puts "Call stream_to_file for #{path}"
    stream_to_file(event.fetch("stream"), path: path)
  end
  client.command("Tracing.end")
end
```
```
Ferrum::Browser
  tracing
    handle tracing.Complete event once
Call stream_to_file for /home/mifrill/Documents/source/rubycdp/ferrum/spec/tmp/trace.json
Call stream_to_file for /home/mifrill/Documents/source/rubycdp/ferrum/spec/tmp/trace2.json
Call stream_to_file for /home/mifrill/Documents/source/rubycdp/ferrum/spec/tmp/trace2.json
Call stream_to_file for /home/mifrill/Documents/source/rubycdp/ferrum/spec/tmp/trace3.json
Call stream_to_file for /home/mifrill/Documents/source/rubycdp/ferrum/spec/tmp/trace3.json
Call stream_to_file for /home/mifrill/Documents/source/rubycdp/ferrum/spec/tmp/trace3.json
      outputs a trace
```
```ruby
def stop
  client.on("Tracing.tracingComplete", handler_type: :once) do |event|
    puts "Call stream_to_file for #{path}"
    stream_to_file(event.fetch("stream"), path: path)
  end
  client.command("Tracing.end")
end
```
```
Ferrum::Browser
  tracing
    handle tracing.Complete event once
call stream_to_file for /home/mifrill/Documents/source/rubycdp/ferrum/spec/tmp/trace.json
call stream_to_file for /home/mifrill/Documents/source/rubycdp/ferrum/spec/tmp/trace2.json
call stream_to_file for /home/mifrill/Documents/source/rubycdp/ferrum/spec/tmp/trace3.json
```

route: I think it makes more sense to subscribe to the events on start. Does tracingComplete have any param that it doesn’t have any more data to send?

Mifrill: > Does tracingComplete have any param that it doesn’t have any more data to send?

nope:
```ruby
Ferrum::Browser
  tracing
Call stream_to_file for /home/mifrill/Documents/source/rubycdp/ferrum/spec/tmp/trace.json
{"dataLossOccurred"=>false, "stream"=>"1", "traceFormat"=>"json", "streamCompression"=>"none"}
Call stream_to_file for /home/mifrill/Documents/source/rubycdp/ferrum/spec/tmp/trace2.json
{"dataLossOccurred"=>false, "stream"=>"2", "traceFormat"=>"json", "streamCompression"=>"none"}
Call stream_to_file for /home/mifrill/Documents/source/rubycdp/ferrum/spec/tmp/trace2.json
{"dataLossOccurred"=>false, "stream"=>"2", "traceFormat"=>"json", "streamCompression"=>"none"}
Call stream_to_file for /home/mifrill/Documents/source/rubycdp/ferrum/spec/tmp/trace3.json
{"dataLossOccurred"=>false, "stream"=>"3", "traceFormat"=>"json", "streamCompression"=>"none"}
Call stream_to_file for /home/mifrill/Documents/source/rubycdp/ferrum/spec/tmp/trace3.json
{"dataLossOccurred"=>false, "stream"=>"3", "traceFormat"=>"json", "streamCompression"=>"none"}
Call stream_to_file for /home/mifrill/Documents/source/rubycdp/ferrum/spec/tmp/trace3.json
{"dataLossOccurred"=>false, "stream"=>"3", "traceFormat"=>"json", "streamCompression"=>"none"}
```
But we can use use `index` to play with `once` type events! :smiley: I didn't catch it, thank you

> I think it makes more sense to subscribe to the events on start.

yeah, it makes sense

README.md
```diff
@@ -1102,6 +1103,33 @@ frame.at_css("//a[text() = 'Log in']") # => Node
 #### select
 
 
+## Tracing
+
+You can use `tracing.start` and `tracing.stop` to create a trace file which can be opened in Chrome DevTools or [timeline viewer](https://chromedevtools.github.io/timeline-viewer/).
+
+```ruby
+browser.tracing.start(path: "trace.json")
```
route, line 1111: How about tracing block version?

```ruby
browser.tracing.record(path: "trace.json") do
  browser.go_to("https://www.google.com")
end
```

README.md
```diff
@@ -1102,6 +1103,33 @@ frame.at_css("//a[text() = 'Log in']") # => Node
...
+browser.go_to("https://www.google.com");
```
route, line 1112: Remove `;` please, if tracing is implemented in the page it's better to use page here rather than browser. I'm gonna rewrite README soon to make it clear.

README.md
```diff
@@ -1102,6 +1103,33 @@ frame.at_css("//a[text() = 'Log in']") # => Node
...
+browser.tracing.stop
+```
+
+#### tracing.start(trace_params:, \*\*options) : `Hash` | `Hash`
+
+* trace_params `Hash`
```
route, line 1118: That's already clear that these are params for a trace, `trace_params` is redundant

README.md
```diff
@@ -1102,6 +1103,33 @@ frame.at_css("//a[text() = 'Log in']") # => Node
...
+
+* trace_params `Hash` 
+  * Specify [Tracing.start PARAMETERS](https://chromedevtools.github.io/devtools-protocol/tot/Tracing/#method-start).
+
+* options `Hash`
+  * `:path` (String) - `String` to save a Trace data output on the disk, not specified by default.
+  * `:encoding` (Symbol) - `:base64` | `:binary` setting only for memory Trace data output, `:binary` by default.
+  * `:screenshots` (Boolean) - When true - Captures screenshots in the trace, `false` by default.
+
+Only one trace can be active at a time per browser.
```
route, line 1126: So it's only one trace per browser but we delegate trace method to a page?

lib/ferrum/page/screenshot.rb
```diff
@@ -42,12 +42,7 @@ def pdf(**opts)
         path, encoding = common_options(**opts)
         options = pdf_options(**opts).merge(transferMode: "ReturnAsStream")
         handle = command("Page.printToPDF", **options).fetch("stream")
-
-        if path
-          stream_to_file(handle, path: path)
-        else
-          stream_to_memory(handle, encoding: encoding)
-        end
+        Utils::Stream.for(self).fetch(handle, encoding: encoding, path: path)
```
route, line 45: These classes should be loosely coupled, and it is definitely not a good idea to pass instance of page to utils. We better instantiate utils stream and yield IO with the block so that we don't pass page and still loop through the stream:

```ruby
Utils::Stream.fetch(handle, encoding: encoding, path: path) do
  command("IO.read", handle: handle, size: STREAM_CHUNK)
end
```

lib/ferrum/page/tracing.rb
```diff
@@ -0,0 +1,82 @@
...
+      end
+
+      def stop
+        client.command("Tracing.end")
+        promise.value!
+      end
+
+      private
+
+      attr_accessor :client, :options, :promise
```
route, line 45: Move it above `initialize`

lib/ferrum/utils/stream.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+module Ferrum
+  module Utils
+    module Stream
+      module_function
+
+      def for(source)
```
route, line 8: That's very very bad to reassign things like that in a module like this

lib/ferrum/page/tracing.rb
```diff
@@ -0,0 +1,82 @@
...
+      EXCLUDED_CATEGORIES = %w[
+        *
+      ].freeze
+
+      def initialize(client:)
+        self.client = client
+      end
+
+      def start(trace_params: {}, **options)
+        self.options = {
```
route, line 28: It's mixed up right with optiosn and params and it's hard to judge what goes where. Need to think about it

lib/ferrum/page/tracing.rb
```diff
@@ -0,0 +1,82 @@
...
+          **options
+        }
+        self.promise = Concurrent::Promises.resolvable_future
+        subscribe_on_tracing_event
+        inner_start(trace_params)
+      end
+
+      def stop
+        client.command("Tracing.end")
+        promise.value!
```
route, line 40: Does it have timeout?

route (commented):
Looks good overall, but we need to polish some things to make it perfect. That's it for the first round! Thank you for the hard work!

README.md
Mifrill: yeah, that's a good point, actually, my intention was to allow customize any racing params that are passed to `Tracing.start` command, I mean those:
Selection_996

In general, I think we can keep only our own params because if someone really needs the tiny trace-params customization he is always able to implement his own tracing class for this, so let's drop `trace_params`

lib/ferrum/page/tracing.rb
Mifrill: do you mean this?
```ruby
module Ferrum
  class Page
    class Tracing
      attr_accessor :client, :options, :promise

      def initialize(client:)
        self.client = client
      end
```
shall we say, it's encapsulation point because the order makes sense to keep methods `private`:
```ruby
irb(main):001:1* class Test
irb(main):002:1*   attr_accessor :string
irb(main):003:1*   
irb(main):004:1*   private
irb(main):005:1*   
irb(main):006:1*   attr_accessor :private_string
irb(main):007:0> end
=> [:private_string, :private_string=]
irb(main):008:0> Test.new.string = "public"
=> "public"
irb(main):009:0> Test.new.private_string = "public"
(irb):9:in `<main>': private method `private_string=' called for #<Test:0x0000558fe07472a8> (NoMethodError)
Did you mean?  private_methods
	from /home/mifrill/.rbenv/versions/3.0.2/lib/ruby/gems/3.0.0/gems/irb-1.3.5/exe/irb:11:in `<top (required)>'
	from /home/mifrill/.rbenv/versions/3.0.2/bin/irb:23:in `load'
	from /home/mifrill/.rbenv/versions/3.0.2/bin/irb:23:in `<main>'
```

lib/ferrum/page/tracing.rb
Mifrill: yup, see this thread: https://github.com/rubycdp/ferrum/pull/189#discussion_r766151731
let's drop `trace_params` for now

lib/ferrum/page/tracing.rb
Mifrill: ambiguous question because the answer is: yes and no at the same time :smile:

no - if the event is not triggered we gonna wait forever in this particular implementation.

yes - we are able to add one more param and pass it this way: `promise.value!(options[:timeout])` if we wanna stop the process when no event is triggered.
Source: https://github.com/ruby-concurrency/concurrent-ruby/blob/5b37a206e88a71939c76cd74854ab223edd56ab8/lib/concurrent-ruby/concurrent/promises.rb#L981-L992

As I understand, a file with a trace may take a long time to get in storage (due to the amount of data), that's why I think it's good to wait till the process is finished completely.

lib/ferrum/utils/stream.rb
Mifrill: okay, good to know :+1: "if you don't try, you won't know", so let's refactor it.

lib/ferrum/page/tracing.rb
route: I don't see any reason to use private attr_accessor, most likely you will use instance variable instead.

Mifrill: sure, no problem, it makes sense to be consistent with the current codebase style

route: It just in general doesn't make sense, current codebase style is not the reason.

Mifrill: speaking frankly, for me it's not a big deal, however, I think it makes sense to use always methods by `attr_`  instead `@` references due to things like possible extend or code-completion by IDE etc. By the way, it also makes sense to be explicit of what variables should be assignable or not by using `attr_writter` and it also makes sense for me to be strict about what of those variables should be available outside of the class by `private` order usage. Anyway, let's use `@` - no problem at all.

route: They are mutually exclusive in my opinion. If you want to add additional logic to the method, you declare it explicitly with explicit access modifier, if not you can simply use instance variable and last you can use attr methods to declare public access to your state. So I have different opinion, you always use instance variables until you need to provide public access to your internal state. That’s my motto until proven wrong

README.md
Mifrill: I think the timeout error is raising because `Concurrent::Promises.resolvable_future` creates an object with the same object_id, that already has a subscription on the `Tracing.tracingComplete` event from the previous test.

route:
I'm moving now a bit, after I settle down all take a looks at all tickets including this one!

PR #190, opened 2021-08-22
Capture error stacktraces

hmdne:
Ferrum loses some bit of information about JS errors thrown. This PR restores stacktraces.


route:
Thank you!

PR #191, opened 2021-09-01
Feature: Node#select / Node#selected

### LINKED ISSUE #58: Pick option in select
I can't figure out a way to pick an option item in a select. 

Example HTML: 
`<select id="cars" name="cars">
    <option value="volvo">Volvo</option>
    <option value="saab">Saab</option>
    <option value="fiat">Fiat</option>
    <option value="audi">Audi</option>
</select>`

Clicking on the select opens the drop-down list. Using tap { browser.keyboard.type(:Down) } did not work. Option cannot be clicked on, because of being unable to calculate quads.  I could be missing something obvious, and any help would be appreciated.

Mifrill:
Close #58

### DIFF
--- a/README.md
+++ b/README.md
@@ -1048,6 +1048,8 @@ browser.playback_rate # => 2000
 #### property
 #### attribute
 #### evaluate
+#### selected : `Array`
+#### select
 
 
 ## Thread safety ##
--- a/lib/ferrum/node.rb
+++ b/lib/ferrum/node.rb
@@ -139,6 +139,30 @@ module Ferrum
       page.evaluate_func(function, self)
     end
 
+    def select(*values)
+      tap do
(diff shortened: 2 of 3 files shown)


Mifrill:
question:
```ruby

      context "when provided texts of disabled option" do
		before do
		  browser.goto("/ferrum/form")
	    end

		it "1. clears selected options with no exception" do
          expect(browser.at_xpath("//*[@id='form_title']").select(["Other"]).selected).to eq([])
        end

        it "2. picks disabled option with no exception" do
          expect(browser.at_xpath("//*[@id='form_title']").select(["Other"]).selected).to eq(["Other"])
        end
      end
```
which is correct 1 or 2?

route (approved):

PR #207, opened 2021-10-20
Clarify proxy examples

ttilberg:
Just a little readme cleanup to clarify the notes, and make the usage more to the point.

### DIFF
--- a/README.md
+++ b/README.md
@@ -574,11 +574,14 @@ while your code denies but it's too late. The block is mandatory now.
 
 ## Proxy
 
-There's no currently any API related to proxy in Chrome CDP protocol, the only thing you can do now is to pass flags to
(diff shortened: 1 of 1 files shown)


route:
Awesome! I so much like it! I was trying different options and decided to go with this one just because none was left.
If I got you right running a server instead of option `:run` will be turned on by option `:server`?

So all options are flat now?

```ruby
proxy: { host: "1.2.3.4", port: 8080, user: "user", password: "pa$$", server: true|false, bypass: "" }
```

I was thinking that someone would need to run local proxy on some dedicated host/port, even if we set `server: true` we still can pass host/port, and that confused me and I end up with run flag.

ttilberg:
Yes, that's what I was thinking. I'll put together a separate PR to implement and chat; I just didn't want to do it if you had an obvious reason to keep the `server` key separate.

PR #211, opened 2021-10-26
Use cliver from github for tests

route:


zw963:
Hi, do you consider remove cliver gem? anyway, this gem never update since 7 years ago, and i search code on ferrum, 

```
ferrum/browser/options/base.rb:31:              path = Cliver.detect(name) and break(path)
ferrum/browser/command.rb:32:        raise Cliver::Dependency::NotFound.new(NOT_FOUND) unless @path
ferrum/browser/xvfb.rb:14:        Cliver.detect("Xvfb")
ferrum/browser/xvfb.rb:21:        raise Cliver::Dependency::NotFound.new(NOT_FOUND) unless @path
```

is there some special use case for must use cliver gem resolve this?

zw963:
BTW:  when run test on it, i get following warning too.

gems/cliver-0.3.2/lib/core_ext/file.rb:15: warning: method redefined; discarding old absolute_path?

route:
@zw963 Yes I think we can remove it, though we need to first write a simple replacement.

PR #217, opened 2021-11-12
Add basic dev environment

jklina:
This adds a basic [Codespaces](https://docs.github.com/en/codespaces) environment for a one button development environment setup. Tests should be able to be run out of the box as this includes a no sandbox installation of Chrome.


route:
Thank you

PR #236, opened 2022-01-07
Implement wait_for_selector

### LINKED ISSUE #82: Implement wait_for_selector

Mifrill:
closes #82


lib/ferrum/frame/dom.rb
```diff
@@ -36,6 +36,33 @@ def body
         evaluate("document.documentElement.outerHTML")
       end
 
+      def wait_for_selector(css: nil, xpath: nil, timeout: 3000, interval: 100)
+        evaluate_func(%(
```
ttilberg, line 40: To be more consistent with the rest of the gem, would you mind changing this to use a `JS` heredoc? This is helpful because many editors support code highlighting in the language specified:

Original:
image


VS:

Subl:
image

Code:
image


I haven't tried the code out yet, but I'm looking forward to this feature!

lib/ferrum/frame/dom.rb
```diff
@@ -36,6 +36,33 @@ def body
...
+              setTimeout(function () {
+                waitForSelector(resolve, reject);
+              }, interval);
+              attempts++;
+            }
+            return new Promise(function (resolve, reject) {
+              waitForSelector(resolve, reject);
+            });
+          }
+        ), css || xpath, css.nil? && !xpath.nil?, timeout, interval, awaitPromise: true)
```
ttilberg, line 63: I was feeling like this arguments list was a little crowded, and also noticed this function could be incorrectly called without raising an exception: one of `css` or `xpath` is required, but not handled. What do you think about something like this:

```ruby
      def wait_for_selector(css: nil, xpath: nil, timeout: 3000, interval: 100)
        raise ArgumentError.new("Provide :css or :xpath, but not both") if css && xpath
        raise ArgumentError.new("Provide :css or :xpath") unless css || xpath

        evaluate_func(<<~JS, css || xpath, !!xpath, timeout, interval, awaitPromise: true)
          function(selector, isXpath, timeout, interval) {
            var attempts = 0;
```

The guard clauses protect the arguments from being called incorrectly, and also simplify the resolution to `isXpath`, since it can now only be one or the other.

lib/ferrum/frame/dom.rb
Mifrill: it makes sense, will change, thanks!

lib/ferrum/frame/dom.rb
Mifrill: well, it looks like too much for one method to be responsible for selector/XPath and also for validations.
However, I see the confusing point here as well. In this case, I think we should use simple methods for more clear API, like those:
```ruby
      def wait_for_selector(css, **options)
        wait_for_element(css: css, **options)
      end
      def wait_for_xpath(xpath, **options)
        wait_for_element(xpath: xpath, **options)
      end
```
The point to merge two ways of element-find logic was a follow a DRY principle for a common JS function, especially this part:
```js
              var element = isXpath
                ? document.evaluate(selector, document, null, XPathResult.FIRST_ORDERED_NODE_TYPE, null).singleNodeValue
                : document.querySelector(selector);
```
therefore, I think, we can find a way to refactor it with a private function by passing a js function, gimme see closer, will do an update soon.

ttilberg: Nice one, yes, I think this is a better direction.

dsisnero:
Can this be merged?  Do the linters need to be changed or do we want to follow the linter advice?

Mifrill:
Hello @dsisnero this PR is on review, it can't be merged without approval.
The linter offenses are not related to this particular PR, we have dedicated PR to solve it:
https://github.com/rubycdp/ferrum/pull/237 that also on review.

route:
I'm sorry guys, with all this stuff happening I barely find time for open source now, hope it's going to calm down soon.

Nakilon:
If anyone needs a simple substitute of `assert_selector` or `Selenium::WebDriver::Wait.new.until &block` and google leads him here, here is a simple ruby snippet to wait for arbitrary things to return `true`:
```
timeout_true = lambda do |timeout = 1, &block|
  Timeout.timeout timeout do
    block.yield or (sleep 0.1; redo)
  end
end
```

shreeve:
Any chance we could have a reviewer review and accept this? It would be very nice to have in the main Gem.

Thanks!

shreeve:
Would this be simpler?

```ruby
class Ferrum::Browser
  def wait_for(want, wait:nil, step:0.1)
    meth = want.start_with?('/') ? :at_xpath : :at_css
    until node = send(meth, want)
      (wait -= step) > 0 ? sleep(step) : break
    end
    node
  end
end
```

route:
Thanks for this, and sorry it sat unreviewed for so long — that's on me.

I've been digging back into #82 and want to take `wait_for_selector` in a different direction than this PR, so I'm going to close it. Wanted to explain why rather than just closing silently.

The main issue is the API and implementation adds more than needs to. `wait_for_css`, `wait_for_xpath`, and a private `wait_for_selector` are three new names for what's really one operation, and none of them reuse `at_css`/`at_xpath` — the element lookup is reimplemented from scratch as a JS string.

That JS is also more machinery than it needs to be: a hand-rolled recursive `setTimeout` loop with an `attempts`/`max` counter, built through `new Function(stringExpr)()` — an expression turned into a string, wrapped in another string, then reconstituted into a callable inside the page. `evaluate_func` already lets you pass a real function directly, no need for the double indirection.

For reference, Puppeteer's actual `waitForSelector` doesn't poll on a timer by default at all — it sets up a `MutationObserver` in an isolated execution context and resolves as soon as a matching mutation happens, in one CDP round trip, with proper handling for navigation/context teardown. What this PR does is closer to Puppeteer's least-used fallback poller, minus the isolation and the navigation handling.

What I'd rather ship is a plain Ruby loop around the existing, unmodified `at_css`/`at_xpath`:

```ruby
def wait_for_selector(selector, within: nil, timeout: @page.timeout, interval: 0.05)
  meth = selector.start_with?("/") ? :at_xpath : :at_css
  start = Utils::ElapsedTime.monotonic_time
  until (node = send(meth, selector, within: within))
    return nil if Utils::ElapsedTime.timeout?(start, timeout)
    sleep(interval)
  end
  node
end
```

No new JS, no changes to `at_css`/`at_xpath`, and it ends up simpler than both this PR and Puppeteer's isolated-world approach, which fits how thin we want ferrum to stay. Closing in favor of the simpler Ruby version.

PR #243, opened 2022-01-29
Add postData getter for network request

rmarot:
This PR is to add a method on the `Ferrum::Network::Request` object in order to allow the optional POST body to be retrieved (or `nil` when it does not exist).

It looks like unit tests are skipped for this object, but let me know if this requires additional changes in order to be merged.

### DIFF
--- a/lib/ferrum/network/request.rb
+++ b/lib/ferrum/network/request.rb
@@ -49,6 +49,7 @@ module Ferrum
       def post_data
         @request["postData"]
       end
+      alias body post_data
     end
   end
 end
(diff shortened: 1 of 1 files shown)


ttilberg:
My small feedback is that this method should be called `#body`. That would be consistent with HTTP in general, other libraries and be more expected. It’s even in your description of the change ;)

rmarot:
> My small feedback is that this method should be called `#body`. That would be consistent with HTTP in general, other libraries and be more expected. It’s even in your description of the change ;)
> 
> edit: or perhaps, at worst, aliasing it?

Yes, honestly I wasn't sure about this one, but it looks indeed more relevant to me.

ttilberg:
Since the CDP method is `postData` they both have merit. Curious what the bosses will say.

route:
Let's alias it as Tim suggested, and I'll merge it

route:
Thank you

rmarot:
> Thank you

Thanks !

PR #245, opened 2022-02-01
Quote browser path to handle spaces

ttilberg:


ttilberg:
My browser path on MacOS is `"/Applications/Google Chrome.app/Contents/MacOS/Google Chrome"` which was not getting properly escaped when launching to get version.

Additionally, it looks like `command` is only used to get the browser path for printing `--version` and can be simplified.

route:
Correct! Thanks

Issue #169, opened 2021-02-28
Document how to download files


alexanderadam:
I probably just missed this but I cannot find the download documentation.

It's probably related to [the `:save_path` option](https://github.com/rubycdp/ferrum/blob/f0cec32c489d204ebf562fd29781f09b337e7450/lib/ferrum/page.rb#L230) but I cannot find any information about that either.

Do you have any advice?

I'm not sure whether this is relevant but there are at least two relevant cases:

1. Explicit downloads that are triggered within ferrum (i.e. "click on a link and save there result" or "download this (background) image")
2. Implicit downloads that are triggered by the server (i.e. with HTTP header `Content-Disposition: attachment` or HTML links like `<a href="foo" download>…`)

route:
I'll try to add test for that but I think it should just store file in the given folder, basically that's it

alexanderadam:
A test would be great! 👍 I guess I missed something somehow.
Now I'm using an ugly workaround (but at least it's working 😉).

Thank you for looking into this.

marcoschicote:
Hi! 
I'm in need of this too. I need to save an image that is part of a web page. Is it possible to download only that image?

Thanks!

PS: @alexanderadam if you can share the ugly workaround I would appreciate it!

alexanderadam:
> @alexanderadam if you can share the ugly workaround I would appreciate it!

hola Marcos, I'm sure afterwards you'd wish that I didn't tell you :laughing: 

:warning: **Disclaimer: this only works for documents that can be viewed within the browser!** :warning:

But at least it is working in my case. I'm sure that there's a native version in ferrum but I wasn't able to find out how (therefore I created this ticket).

```ruby
require 'mhtml' # for parsing mhtml - https://github.com/benjineering/mhtml_rb

download_directory = 'your/download_directory/'
browser.go_to download_url

# we are storing the current page with all assets within a temporary directory
Dir.mktmpdir do |dir|
  path = "#{dir}/tmp.mhtml"
  browser.mhtml(path: path)
  mhtml = Mhtml::RootDocument.new(File.read(path))

  # extracting all assets from mhtml
  mhtml.sub_docs.each do |file|
    file_name = File.join(download_directory, File.basename(mhtml.sub_docs.first.file_path))
    File.write(file_name, file.body)
  end

browser.back # going back since we don't want to pollute the running session
```

marcoschicote:
Thanks @alexanderadam!
[...]

PR #254, opened 2022-04-14
Add a simple download spec and mention save_path in the readme

nickhammond:
For #169 and to add `save_path` to the readme.

### DIFF
--- a/spec/download_spec.rb
+++ b/spec/download_spec.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+module Ferrum
+  describe Browser do
+    context "download support" do
+      let(:browser) do
+        Ferrum::Browser.new(
+          base_url: Ferrum::Server.server.base_url,
+          save_path: "/tmp/ferrum"
+        )
+      end
+
+      it "saves an attachment" do
+        browser.go_to("/attachment.pdf")
+
+        expect(File.exist?("/tmp/ferrum/attachment.pdf")).to be true
+      end
+    end
+  end
+end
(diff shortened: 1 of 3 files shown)


route:
Thank you!

spec/download_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+          base_url: Ferrum::Server.server.base_url,
+          save_path: "/tmp/ferrum"
+        )
+      end
+
+      it "saves an attachment" do
+        browser.go_to("/attachment.pdf")
+
+        expect(File.exist?("/tmp/ferrum/attachment.pdf")).to be true
+      end
```
Mifrill, line 17: :thinking: weired point is this test failing for me locally:
```
Failures:

  1) Ferrum::Browser download support saves an attachment
     Failure/Error: expect(File.exist?("/tmp/ferrum/attachment.pdf")).to be true
     
       expected true
            got false
     # ./spec/download_spec.rb:16:in `block (3 levels) in <module:Ferrum>'

Finished in 0.49189 seconds (files took 0.57409 seconds to load)
1 example, 1 failure
```
For both ruby versions: 2.7.4 and for 3.0.2.

route: Yea I'll fix

nickhammond: After I prepared this I noticed that Ferrum was really consistent when not using the `url` option but once I used the `url` option with a docker container(browserless/chrome) then it was really inconsistent with downloads and would often hang. 

@Mifrill Any indicators as to why it's failing?

route: @nickhammond  try master branch, is it any better now?

nickhammond: @route I can't seem to get the same behavior from yesterday but it's consistently working when downloading something so should be good. I have a failing spec locally but it's related to window position "allows the window to be positioned" browser_spec:122 so not the same as on the above but no worries.

route:
Yea me too, looks like something broke on Chrome's side... needs investigation

spec/download_spec.rb
route: Yea me too, looks like something broke on Chrome's side... needs investigation

Issue #253, opened 2022-04-09
Improve the flow of retrieving and setting Cookies


gbaptista:
Today, I have to set cookies through:

```ruby
browser.cookies.set(name: "stealth", value: "omg", domain: "google.com")
```

This is great, but if I want, for example, to use a retrieved cookie:

```ruby
some_cookie = browser.cookies.all.values.first
# #<Ferrum::Cookies::Cookie:0x00007f96b038e6b0
#  @attributes=
#   {"name"=>"stealth",
#    "value"=>"omg",
#    "domain"=>"google.com"}>

browser.cookies.set(some_cookie)
# => ArgumentError: wrong number of arguments (given 1, expected 0)
```

To make the cookie object comply with the expected API, I need to workaround:

```ruby
adapted_cookie = {}

some_cookie.instance_variable_get('@attributes').each do |key, value|
  cookie[key.to_sym] = value
end

browser.cookies.set(**adapted_cookie)
```

Why would that be important? For my use case, I store cookies and then retrieve them at another moment in the future:

```ruby
# Save:
File.write('cookies.bin', Marshal.dump(browser.cookies.all))

# Load:
Marshal.load(File.read('cookies.bin')).values.each do |raw_cookie|
  adapted_cookie = {}
  raw_cookie.instance_variable_get('@attributes').each do |key, value|
    adapted_cookie[key.to_sym] = value
  end
  browser.cookies.set(**adapted_cookie)
end
```

Some ideas to improve this flow:

- Expose the `@attributes`, so I don't need to `instance_variable_get`;
- Create a `to_h` method to have access to a hash accepted in `cookies.set`;
- Just allow `cookies.set` to accept instances of `Ferrum::Cookies::Cookie`.

route:
Totally agree it's definitely not thoroughly thought now and needs improvement.

gbaptista:
Closed by:
- #256

PR #256, opened 2022-04-14
Improving the flow of retrieving and setting Cookies

gbaptista:
Related to:
- #253 

Allows `browser.cookies.set` to receive either a hash for setting a new cookie or a previously retrieved cookie.

### DIFF
--- a/lib/ferrum/cookies.rb
+++ b/lib/ferrum/cookies.rb
@@ -3,6 +3,8 @@
 module Ferrum
   class Cookies
     class Cookie
+      attr_reader :attributes
+
       def initialize(attributes)
         @attributes = attributes
       end
@@ -61,10 +63,11 @@ module Ferrum
       all[name]
     end
 
-    def set(name: nil, value: nil, **options)
-      cookie = options.dup
-      cookie[:name]   ||= name
-      cookie[:value]  ||= value
+    def set(options)
+      cookie = (
+        options.is_a?(Cookie) ? options.attributes : options
+      ).dup.transform_keys(&:to_sym)
+
       cookie[:domain] ||= default_domain
 
       cookie[:httpOnly] = cookie.delete(:httponly) if cookie.key?(:httponly)
--- a/README.md
+++ b/README.md
@@ -733,6 +733,18 @@ Sets given values as cookie
 browser.cookies.set(name: "stealth", value: "omg", domain: "google.com") # => true
 ```
 
+#### set(cookie) : `Boolean`
+
+Sets given cookie
+
+* cookie `Cookie`
+
+```ruby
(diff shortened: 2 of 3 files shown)


lib/ferrum/cookies.rb
```diff
@@ -61,10 +63,11 @@ def [](name)
...
     end
 
-    def set(name: nil, value: nil, **options)
-      cookie = options.dup
-      cookie[:name]   ||= name
-      cookie[:value]  ||= value
+    def set(options)
+      cookie = (
+        options.is_a?(Cookie) ? options.attributes : options
+      ).dup.transform_keys(&:to_sym)
```
Mifrill, line 69: `transform_keys` looks redundant
```suggestion
      ).dup
```

Mifrill (approved):
Look great! :+1:

lib/ferrum/cookies.rb
gbaptista: @Mifrill Thanks for the suggestion!

I did some experiments, and while this looks redundant, it's needed and makes me realize that I'm not testing the feature enough.

The stored attributes will use string keys when cookies are defined through the browser. So, the following use cases would fail:
```ruby
browser.cookies.set(cookie)

browser.cookies.set(cookie.attributes)
```

For example, the `domain` would be overwritten by `cookie[:domain] ||= default_domain` as `cookie[:domain]` != `cookie["domain"]`.

I updated the PR, including more tests, to make it more explicit and also ensure the following expected behaviors:

- We do not create side effects by mutating the options Hash: [L136](https://github.com/rubycdp/ferrum/pull/256/files#diff-a880348c9c62b3e07cb738dd7539f6a10a9ebd0a2af872c9db2d84777800791eR136)
- We do not create side effects by mutating the cookie Object. [146](https://github.com/rubycdp/ferrum/pull/256/files#diff-a880348c9c62b3e07cb738dd7539f6a10a9ebd0a2af872c9db2d84777800791eR146)
- We ensure all the characteristics of the reference cookie. [L106](https://github.com/rubycdp/ferrum/pull/256/files#diff-a880348c9c62b3e07cb738dd7539f6a10a9ebd0a2af872c9db2d84777800791eR106)
- We ensure all the characteristics of the reference attributes: [L129](https://github.com/rubycdp/ferrum/pull/256/files#diff-a880348c9c62b3e07cb738dd7539f6a10a9ebd0a2af872c9db2d84777800791eR129)

That way, if we remove either `dup` or `transform_keys(&:to_sym)`, the tests will fail.

route:
Thank you!

lib/ferrum/cookies.rb
Mifrill: alright, thanks for thorough explanation :+1:

PR #257, opened 2022-04-14
Consider renaming the default branch from master

gbaptista:
I am making the necessary changes to help and encourage the renaming process.

More context and a step-by-step at https://github.com/github/renaming

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -2,7 +2,7 @@ name: Tests
 on:
   push:
     branches:
-      - master
+      - main
   pull_request:
 
 jobs:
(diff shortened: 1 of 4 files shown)


Mifrill (approved):
> We like it because it's short, it keeps your muscle memory intact, and it translates well across most languages.

Source: https://github.com/github/renaming#why-main
TIL. :smile:

route:
Thanks

Nakilon:
This is an awful change. Just that somewhere in the country that owns the Github they have social problem and hate each other they decide to make it difficult for the whole world just to laugh at people watching them having issues with this change.

PR #258, opened 2022-04-15
Improving browser options semantics

gbaptista:
Today, to set a browser option, you need to:

```ruby
Ferrum::Browser.new(browser_options: { 'no-sandbox': nil })
```

This semantics can be misleading because, at least from my perspective, `nil` is associated with "absence."

It may also mislead people into trying to remove a default option by setting it as `nil`:

```ruby
Ferrum::Browser.new(browser_options: { 'disable-extensions': nil })
```

This PR proposes the use of `true` for desired flags:

```ruby
Ferrum::Browser.new(browser_options: { 'no-sandbox': true })
```

And the use of `nil` or `false` to allow the possibility of easily removing a default option:

```ruby
Ferrum::Browser.new(browser_options: { 'disable-extensions': false })
Ferrum::Browser.new(browser_options: { 'disable-extensions': nil })
```

There's also protection to prevent people from removing required flags:
```ruby
Ferrum::Browser.new(browser_options: { 'remote-debugging-port' => nil })
# => #<ArgumentError: remote-debugging-port is required>
```

I choose an explicit error if the person explicitly tries to remove a required flag rather than just overriding it with the required value. It could confuse the person not understanding why the removal didn't work.

Also, README was updated to instruct the new semantics.

**Overall**

Upsides:
- A more intuitive way to work with flags.
- Possibility to easier remove default flags.
- Protection against removing required flags.
- Potentially fewer issues and frustration compared to the previous way.

Downsides:
- As today, people use `nil` to define a flag, so we may consider this a potential breaking change.

### DIFF
--- a/lib/ferrum/browser/options/chrome.rb
+++ b/lib/ferrum/browser/options/chrome.rb
@@ -5,41 +5,41 @@ module Ferrum
     module Options
       class Chrome < Base
         DEFAULT_OPTIONS = {
-          "headless" => nil,
-          "disable-gpu" => nil,
-          "hide-scrollbars" => nil,
-          "mute-audio" => nil,
-          "enable-automation" => nil,
-          "disable-web-security" => nil,
-          "disable-session-crashed-bubble" => nil,
-          "disable-breakpad" => nil,
-          "disable-sync" => nil,
-          "no-first-run" => nil,
-          "use-mock-keychain" => nil,
-          "keep-alive-for-test" => nil,
-          "disable-popup-blocking" => nil,
-          "disable-extensions" => nil,
-          "disable-hang-monitor" => nil,
+          "headless" => true,
+          "disable-gpu" => true,
+          "hide-scrollbars" => true,
+          "mute-audio" => true,
+          "enable-automation" => true,
+          "disable-web-security" => true,
+          "disable-session-crashed-bubble" => true,
+          "disable-breakpad" => true,
+          "disable-sync" => true,
+          "no-first-run" => true,
+          "use-mock-keychain" => true,
+          "keep-alive-for-test" => true,
+          "disable-popup-blocking" => true,
+          "disable-extensions" => true,
+          "disable-hang-monitor" => true,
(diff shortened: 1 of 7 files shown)


lib/ferrum/browser/options/chrome.rb
```diff
@@ -63,7 +63,23 @@ def merge_required(flags, options, user_data_dir)
...
+          %w[remote-debugging-port remote-debugging-address
+             window-size user-data-dir].each do |required|
+            next unless options[:browser_options].key?(required)
+
+            if options[:browser_options][required] == false ||
+               options[:browser_options][required].nil?
+              raise ArgumentError, "#{required} is required"
+            end
+          end
+        end
```
Mifrill, line 78: This method is a good candidate to be in `lib/ferrum/browser/options/base.rb` to follow the DRY principle.

gbaptista: @Mifrill Agree. I made the change and updated the PR to avoid this repetition. Thanks for pointing it out!

Mifrill (approved):
LGTM!

route:
I'm afraid it changes existing behaviour and there's no a single deprecation warning, I'll get back to it later.

route:
Since it's a breaking change and there are no deprecation warnings. It's not a suitable solution for now.
For a command line flag which in general can be `--flag` `--flag=value` or `-flag` `-flag value` hash fits just well.
Having a key with nil value mirrors `--flag`, having a key with some value mirrors `--flag=value`, absence of k,v is absence of flag. It's counter-intuitive to pass nil, false and expect a flag to be removed instead of `--flag=false` in my opinion.

We should keep using current approach especially when it's been in use for years, or replace hash with something more complex, it needs research, but maybe something like this:

```ruby
  options = Chrome::Options.new
  options.add_argument('--ignore-certificate-errors')
  options.add_argument('--disable-popup-blocking')
  options.add_argument('--disable-translate')
```

PR #264, opened 2022-05-08
Fix #158 Switch browser to offline mode

route:


twe4ked:
Also closes https://github.com/rubycdp/cuprite/issues/189!

PR #266, opened 2022-05-12
Supported Ruby 2.6

route:


route:
To make everyone happy @Nakilon I returned back 2.6 but I will discontinue using it later this year, don't want to spend time supporting it.

Nakilon:
Thank you!

PR #269, opened 2022-05-25
Add `env` option

BuonOmo:
**Issue**

Existing env variables may have an impact on a given chrome instance, and they sometimes overlap with env variables that are used in ruby. 

For instance, we've been having trouble with our dynamically linked jemalloc that is exported our env (`LD_PRELOAD`). The chrome instance didn't like this allocator version.

**Solution**

Hence I believe it would be nice to give one the opportunity to set those _only for the subprocess_.


route:
@BuonOmo but how this would save you from the issue you described? If you set LD_PRELOAD for ruby it will be inherited by Chrome right?

BuonOmo:
@route per [ruby documentation](https://devdocs.io/ruby~3/process#method-c-spawn):

> If a hash is given as env, the environment is updated by env before `exec(2)` in the child process. If a pair in env has nil as the value, the variable is deleted.
> ```ruby
> # set FOO as BAR and unset BAZ.
> pid = spawn({"FOO"=>"BAR", "BAZ"=>nil}, command)
> ```

So here: `{ "LD_PRELOAD" => nil }` would do the trick.

PS: I see that my test is misleading, fixing that

PR #281, opened 2022-08-10
Handle file not found error in `Page#go_to`

y-yagi:
Chrome allows to specify file scheme to an URL.
When a non-existent file is specified, Chrome returns `ERR_CONNECTION_TIMED_OUT`. But `ferrum` doesn't handle that error.
So in that case, an error happened, but it looked to work correctly.

This fixed to handle that error inside `go_to` and correctly raise an error.


route:
Thank you!

PR #293, opened 2022-10-01
Added `Ferrum::Browser#version`

postmodern:
* Added `Ferrum::Browser#version` which calls the [`Browser.getVersion`](https://chromedevtools.github.io/devtools-protocol/1-3/Browser/#method-getVersion) command.
* Also added `Ferrum::Browser::VersionInfo` to wrap around the returned JSON hash.


route:
Looks good, but the default style for quotes in this project is `"` and we can merge!

postmodern:
Amended.

route:
Oh I forgot, we need to add to changelog a couple of entries as well, but we can do it in the next PR

postmodern:
See PR #295.

Issue #197, opened 2021-09-28
Exchange#finished returns true before response body is ready


ttilberg:
Hello, I'm trying to intercept the response body of an XHR request. After looking through the spec files, I saw that Ferrum keeps track of all of the network traffic, which is convenient for this case. (I tried using on('Network.responseReceived') hooks, but couldn't work it out.)

While waiting for the correct `Exchange` to be `finished?` I kept running into `Ferrum::BrowserError: No data found for resource with given identifier`.  Example code:

```rb
    response = Timeout.timeout(15) do
      # Find ajax request for search results
      until xhr = browser.network.traffic.find { |exchange| exchange.request.url =~ /search\/bySize/ }
        print '.'
        sleep 0.2
      end
      # Wait for response to complete
      until xhr.finished?
        print 'x'
        sleep 0.2
      end
      xhr.response.body
    end
```

After some research, I've learned that instead you should wait for `Network.loadingFinished` before trying to get the response body. 

I wanted to highlight a few notes in Ferrum's API as I wonder if `Exchange#finished?` should wait for that event, or if the `Response` object c/should be more clever. (Or both).

```rb
# network.rb #subscribe
# ...
      @page.on("Network.responseReceived") do |params|
        if exchange = select(params["requestId"]).last
          response = Network::Response.new(@page, params)
          exchange.response = response
        end
      end

      @page.on("Network.loadingFinished") do |params|
        exchange = select(params["requestId"]).last
        if exchange && exchange.response
          exchange.response.body_size = params["encodedDataLength"]
        end
      end
```

```rb
# exchange.rb
# ...
      def finished?
        blocked? || response || error
      end
```

Note how the exchange is `finished?` when it has a `response` object. But also notice that this response attribute is assigned in the `Network.responseReceived` hook, not the `Network.loadingFinished` hook. There is, however, a property that gets set at `Network.loadingFinished`, `body_size`, so I tried keying off that:

```rb
    response = Timeout.timeout(15) do
      # Find ajax request for search results
      until xhr = browser.network.traffic.find { |exchange| exchange.request.url =~ /search\/bySize/ }
        print '.'
        sleep 0.2
      end
      # Wait for response to complete
      until xhr&.response&.body_size
        print 'x'
        sleep 0.2
      end
      xhr.response.body
    end
```

With this change, this code no longer raises the noted exception, though the API is being used in a strange way.

[...]

PR #294, opened 2022-10-01
Add support for `browser.on(:response)`.

postmodern:
I wanted to be able to have a callback similar to `on(:request)` but triggered once a response was received. This adds an additional callback for the [`Network.responseReceived`](https://chromedevtools.github.io/devtools-protocol/1-3/Network/#event-responseReceived) event when a response is received. It will look up the `Ferrum::Network::Exchange` object with the matching `requestId` and passes it back to the block, once a response has been received for a request.

I didn't know where to add specs for this, since there does not appear to be specs for `Ferrum::Page` or a `spec/page_spec.rb` file? I also didn't know where I should mention `browser.on(:response)` in the README, since `browser.on(...)` is only mentioned in conjunction to using `browser.network.intercept`.

### DIFF
--- a/lib/ferrum/page.rb
+++ b/lib/ferrum/page.rb
@@ -303,6 +303,14 @@ module Ferrum
           exchange.intercepted_request = request
           block.call(request, index, total)
         end
+      when :response
+        @client.on('Network.responseReceived') do |params, index, total|
+          exchange = network.select(params["requestId"]).last
+
+          if exchange
+            block.call(exchange, index, total)
+          end
+        end
       when :auth
         @client.on("Fetch.authRequired") do |params, index, total|
           request = Network::AuthRequest.new(self, params)


postmodern:
@route would appreciate some guidance on this new feature and where I should add the specs and documentation?

route:
@postmodern I was reorganizing specs so currently page specs are in the browser_spec. I think we can create page_spec and put it there. Though it overlaps with network_spec a bit, because `page.on(:request)` only works with interception, but page.on(:response) works without it. I'm now thinking that maybe it's better to use something like `page.network.on(:request | :response | :auth)`. Need to experiment with that...

postmodern:
@ttilberg hmm that solved the exception, except now it does not appear that my `on(:response)` callback is triggering.

ttilberg:
@postmodern I took a look but didn’t find a CDP event translation helper for `on(:response)`. However, in an earlier issue I noted and used `network.responseReceived`, might this get you on your way? 

https://github.com/rubycdp/ferrum/issues/197#issue-1009903139

postmodern:
@ttilberg yeah it's added in this PR. Feel free to pull my branch and try to mess with it. I swear it was working last year.

postmodern:
OK I added some echo debugging and the `@client.on('Network.responseReceived')` callback is being called, but it's not finding the exchange object for the given `params['requestId']`, so the passed in `block` is never called.

postmodern:
Got it, I needed to do `network.select` from the context of `Ferrum::Page`.

ilyacherevkov:
This is useful, I use it as a monkey patch while it's not merged.

ilyacherevkov:
Turned out it doesn't work for really heavy responses (I m testing particular response which returns 20MB), so I needed to use `Network.requestWillBeSent` to save mapping of "requestId" and 'url', `Network.loadingFinished` to ensure full response got loaded. Then I used mapping table to match requestId with url.

route:
I guess we can go with https://github.com/rubycdp/ferrum/commit/77c8b969a6cbc4146dce41d57d10fb5e2d152344, let's see if it's useful this way.

PR #296, opened 2022-10-03
Add yard

postmodern:
* Added the YARD development dependency and rake task.
* Add YARD documentation to methods by copying existing documentation from the `README`.


route:
🔥

Issue #234, opened 2021-12-27
Way to resolve promise in page javascript?


hkmaly:
Could you add some way to resolve promise? `evaluate('await something_returning_promise()')` obviously doesn't work. `evaluate_async('something_returning_promise()')` doesn't work either (less obviously, returns timeout error). Tried to add method
```
def evaluate_await(expression, *args)
    expression = "function() { return %s }" % expression
    call(expression: expression, arguments: args, awaitPromise: true)
end
```
but, while it seems to do something, it returns `{}`. Also, monkey patching it in is quite complicated.

hkmaly:
Wait, that was my mistake INSIDE the javascript. The evaluate_await works.

route:
https://github.com/rubycdp/ferrum/blob/master/lib/ferrum/frame/runtime.rb#L53

hkmaly:
@route I explicitly said that evaluate_async is not working, always timeouts. Or is there some trick how it's supposed to be used?

ttilberg:
@hkmaly  What is your example? You haven’t provided much information about your original issue for route to consider.

hkmaly:
I have following download method:
```
def download_fallback(browser, url, name, mode)
    return(false) if File.exists?(name)
    # https://stackoverflow.com/questions/44698967/requesting-blob-images-and-transforming-to-base64-with-fetch-api/5046
3054#50463054
    js = <<~ZATREOT
fetch('#{url}').then(response => response.blob()).then(blob => new Promise(callback => {
    let reader = new FileReader();
    reader.onload = function() {callback(this.result)};
    reader.readAsDataURL(blob);}
))
ZATREOT
    if mode
        data = browser.evaluate_await(js);
    else
        data = browser.evaluate_async(js, 5);
    end
    dp = data.split(',')
    return(false) unless dp.count == 2
    data = Base64.decode64(dp[1])
    File.open(name, 'w') {|fd|
        fd.write data
    }
    puts "\tSaved #{name} from #{url}"
    return(true)
rescue => ex
    puts "\tError download #{url}: #{ex.class}: #{ex.message}\n\t#{ex.backtrace.join("\n\t")}"
    return(false)
end
```
In "true" mode, using my own evaluate_await, it works and saves the image from provided url. In "false" mode, using evaluate_async, throws:
```
    Error fallback download (url redacted): Ferrum::ScriptTimeoutError: Timed out waiting for evaluated script to return a value
[...]

PR #325, opened 2023-01-11
Use Promise in example for #evaluate_async

ttilberg:
Related to #234

I couldn't remember how to get a value back from `evaluate_async` and remembered I had created an example in a previous issue. 

The existing example leverages `arguments[0]` and calls a simple object. Instead, I would like to see an actual, practical example using familiar things that people would commonly use this for.

Additionally, I think we should encourage usage of `__f()` as in the next example, instead of `arguments[0]`, or rework the `#evaluate_async` template. As an example, if you call:

```ruby
browser.evaluate_async '__f(arguments)', 5, "arg 1", 2, 3
# => ["arg 1", 2, 3, {}]
```

`arguments[0]` no longer refers to the fulfill callback. You end up receiving ` TypeError: arguments[0] is not a function (Ferrum::JavaScriptError)`.

Perhaps this highlights some tweaks that could be made to `evaluate_async`, but for now, I'd like to get this docs change up asap.

### DIFF
--- a/README.md
+++ b/README.md
@@ -826,15 +826,19 @@ browser.evaluate("[window.scrollX, window.scrollY]")
 
 #### evaluate_async(expression, wait_time, \*args)
 
-Evaluate asynchronous expression and return result
+Evaluate an asynchronous expression and return result
 
-* expression `String` should be valid JavaScript
+* expression `String` should be valid JavaScript, and include the fulfill callback: `__f()`
 * wait_time How long we should wait for Promise to resolve or reject
 * args `Object` you can pass arguments, though it should be a valid `Node` or a
 simple value.
 
 ```ruby
-browser.evaluate_async(%(arguments[0]({foo: "bar"})), 5) # => { "foo" => "bar" }
+browser.evaluate_async <<JS, 5
+  fetch("https://www.httpbin.org/json")
+    .then(response => response.json())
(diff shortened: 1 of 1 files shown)


route:
I don't like current API either, but I remember picking up the names for `resolve` and `reject` functions so that they don't collide with any other name (thus I added `__`), and now we are making them kinda explicit. Looking at:

```ruby
browser.evaluate_async '__f(arguments)', 5, "arg 1", 2, 3
```

is so unreadable. I'm always confused when to put what, and use specs to remember how to use it. So maybe it's time to improve the API then if it's doable at all? :)

ttilberg:
I ended up here to reference the `__r` trick again. @route I noticed you have a new issue from Jan for planning a 1.0 release that may include this type of change. I was thinking today "what if the resolve argument was called `resolveToFerrum`" or something. Putting `ferrum` in the name would help isolation as you wanted, but keep meaningful language that helps people connect the dots. Just wanted to put the thought forward for consideration when you get around to it.

route:
@ttilberg yea we can do that, I was just thinking maybe there's a way to make it even more convenient back then. No considerations taken so far

route:
Going to implement new API here https://github.com/rubycdp/ferrum/pull/617

PR #331, opened 2023-01-22
Limit emulation to enabling slowmoable

### LINKED ISSUE #330: Emulation.setDeviceMetricsOverride causes wrong scale to be used for screenshots
On a high DPI screen, where window.devicePixelRatio is 2, for example, taking screenshots should use a scale of 2. It would, by default if it weren't for the fact that `Emulation.setDeviceMetricsOverride` is invoked, setting the `deviceScaleFactor` to 1. It's unclear why this is necessary, as both headed and headless mode appear to work without it.

Furthermore, if you take a screenshot during headed mode and `deviceScaleFactor` is reduced to 1, it causes an unsightly flash as Chrome scales the viewport down by half before taking the screenshot.

It also seems to be related to and/or cause this issue: https://github.com/rubycdp/ferrum/issues/203

Is the line necessary at all or can it be removed?

Thanks!

aaronjensen:
Fixes #330 

If the line cannot be removed (because `slowmoable` needs to stay enabled) then this will disable the override of everything else.

### DIFF
--- a/lib/ferrum/page.rb
+++ b/lib/ferrum/page.rb
@@ -145,7 +145,7 @@ module Ferrum
       command("Emulation.setDeviceMetricsOverride", slowmoable: true,
                                                     width: width,
                                                     height: height,
-                                                    deviceScaleFactor: 1,
+                                                    deviceScaleFactor: 0,
                                                     mobile: false,
                                                     fitWindow: false)
     end


aaronjensen:
@route would you be able to take a look at this please?

route:
@aaronjensen could you please rebase it with master? I don't see how to run tests for this PR

aaronjensen:
Rebased. When you say you don't see how to run tests, are you saying that you don't see any tests or you this PR somehow breaks the normal means of running tests? The emulation device metric overrides were not previously tested afaict, so I wasn't sure how to add a test for them.

aaronjensen:
I do see a test failure actually in browser_spec.rb, specifically:

```
include_examples "resize viewport by fullscreen" do
  let(:path) { "/ferrum/custom_html_size_100%" }
  let(:viewport_size) { [1272, 1008] }
end
```

route:
@aaronjensen I was talking about a button to "Approve and Run tests" for this PR. Now it appeared. We need to investigate this failure. Cannot merge it with a breaking test ;)

aaronjensen:
Updated to only disable overriding the device scale factor. Setting the height and width still work and are necessary to allow controlling height and width when using full screen.

route:
Looks good now!

aaronjensen:
Thank you!

aaronjensen:
@route sorry to bother you again, but do you have an idea when you plan to release a new gem version with this change? Thank you!

route:
I don’t know, when it’s time :) use master

aaronjensen:
Understood, thanks for the response.

PR #341, opened 2023-02-15
Raise `Ferrum::StatusError` for any top frame navigation error

francisbeaudoin:
## Context
The `Page#go_to` is only raising a `Ferrum::StatusError` for a subset of exceptions:

https://github.com/rubycdp/ferrum/blob/0cab51d1f0500efa7573cac39f99c44421f81c86/lib/ferrum/page.rb#L113-L124

Per the Chrome DevTools [documentation](https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-navigate) an `errorText` is present if and only if navigation has failed.

## Proposal
This PR is proposing to raise regardless of the `errorText` given that the navigation failed. 

> **Warning**
This could be considered as a breaking change.

## Reference
Discussion: https://github.com/rubycdp/ferrum/discussions/340

### DIFF
--- a/README.md
+++ b/README.md
@@ -595,9 +595,8 @@ Activates offline mode for a page.
 
 ```ruby
 browser.network.offline_mode
-browser.go_to("https://github.com/") # => Ferrum::StatusError (Request to https://github.com/ failed to reach server, check DNS and server status)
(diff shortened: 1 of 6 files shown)


route (approved):

route:
Thank you! Awesome!

PR #367, opened 2023-07-05
Add `Ferrum::Cookies#each`

postmodern:
* Added `Ferrum::Cookies::Cookie#each`.
* Also make `Ferrum::Cookies` `Enumerable`.
* Refactored `Ferrum::Cookies#all` to use `#each`.
* Optimize `Ferrum::Cookies#[]` to use `find { |cookie| ... }`.

### DIFF
--- a/lib/ferrum/cookies.rb
+++ b/lib/ferrum/cookies.rb
@@ -69,7 +69,7 @@ module Ferrum
     #   # }>
     #
     def [](name)
-      all[name]
+      find { |cookie| cookie.name == name }
     end
 
     #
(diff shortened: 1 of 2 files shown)


postmodern:
The motivation for this PR is to allow selecting cookies by with the same `name` but different `Path` or `Domain` values. `cookies.all` would combine all cookies into a Hash, so cookies with the same `name` but different `Domain` would overwrite each other.

route (approved):

route:
Perfect! Thank you!

Issue #352, opened 2023-04-12
Use `--headless=new` Chrome flag


singhprd:
Hello! 

I have just read this: https://developer.chrome.com/articles/new-headless/

TLDR: 

> Chrome headless was released in 2017, and was technically [a separate, alternate browser implementation](https://source.chromium.org/chromium/chromium/src/+/main:headless/;drc=c67febd82ae3e18ac8db1397f4ccfa87b0da2ffc) that happened to be shipped as part of the same Chrome binary.
> 
> The Chrome team are unifying Headless and headful modes
> 
> To use the new mode start chrome with `--headless=new` 
> 
> Currently, passing the --headless command-line flag without an explicit value still activates the old Headless mode — Chrome plan to change this default to new Headless over time.

I was wondering about starting a discussion here -  should Ferrum default to setting `--headless=new`?

There might be more to do that just update the option here so it'd be great to hear thoughts?:
https://github.com/rubycdp/ferrum/blob/main/lib/ferrum/browser/options/chrome.rb#L8

Proof of concept update from my fork here:
https://github.com/fac/ferrum/pull/8/files

route:
Yep, I was looking at this flag for a while, and I think it's good idea to turn it on by default instead of headless=true

route:
@JacobEvelyn I don't see any PR

JacobEvelyn:
Oh sorry; I thought @singhprd's draft PR linked above was in this repo instead of a fork. @singhprd do you want to turn it into a PR here? @route @singhprd would it be helpful if I did?

route:
@JacobEvelyn don't ask, do if you can ;)

route:
As I can see it's not that easy as new Chrome works different with frames, but I started to work on this.

jtoy:
is it merged?

route:
There’s nothing to merge yet

route:
I'm closing this issue as there's a [PR](https://github.com/rubycdp/ferrum/pull/379) you can follow it.

PR #379, opened 2023-07-16
Initial support for headless=new

route:
Addresses #352

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,4 +1,3 @@
 --color
---fail-fast
 --format=doc
 --require spec_helper
--- a/lib/ferrum/browser/options/chrome.rb
+++ b/lib/ferrum/browser/options/chrome.rb
@@ -5,7 +5,7 @@ module Ferrum
     class Options
       class Chrome < Base
         DEFAULT_OPTIONS = {
-          "headless" => nil,
+          "headless" => "new",
           "disable-gpu" => nil,
           "hide-scrollbars" => nil,
           "mute-audio" => nil,
--- a/lib/ferrum/page.rb
+++ b/lib/ferrum/page.rb
@@ -391,7 +391,8 @@ module Ferrum
       resize(width: width, height: height)
 
       response = command("Page.getNavigationHistory")
-      return unless response.dig("entries", 0, "transitionType") != "typed"
+      transition_type = response.dig("entries", 0, "transitionType")
+      return if transition_type == "auto_toplevel"
 
       # If we create page by clicking links, submitting forms and so on it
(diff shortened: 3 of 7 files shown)


lifesaverluke:
Anything we can do to help get this implemented @route?

route:
@lifesaverluke not that I can think of, I'm on it but I was sick and on the road.

route:
It looks like even though I can be able to land support with frames, there's at least one quite major [regression](https://bugs.chromium.org/p/chromium/issues/detail?id=1481047) for Chrome which shows duplicated requests and there's an event missing which is quite crucial for network. So it's not gonna happen until they fix it.

lifesaverluke:
A think that changed for sure is that `last_exchange` is now the request to favicon.ico instead of the page. The old headless mode never loaded favicons, so that change makes sense. 

Maybe we just need to adjust the tests?

route:
> A thing that changed for sure is that `last_exchange` is now the request to favicon.ico instead of the document. The old headless mode never loaded favicons, so that change makes sense.
> 
> Maybe we just need to adjust the tests?

No it’s not that simple, networking is broken

lifesaverluke:
Bit weird that puppeteer / playwright etc do work with headless=new?

route:
@lifesaverluke in general it works, but there are a lot of tests broken for networking. The design is different, the architecture is also different. But looks like I found solution to make it work `disable-site-isolation-trials`

route:
Since "new" mode has a bug with file download behavior it's not default now, but you can turn on "new" mode if it satisfies your needs.

sandstrom:
@route Awesome! 😄 

Will there be a new release later, with this included?

route:
released

janko:
@route I appreciate the addition! Could we also get a new release of Cuprite? The current version locks Ferrum to `0.13.x`, so I cannot upgrade to `0.14.x`.

route:
@janko yep I was going to do that, but need to do some stuff offline ;) life is life)

janko:
@route Of course, as a fellow open source maintainer, I totally understand 😉 I really appreciate all the work you did on Ferrum/Cuprite, it brought us [tremendous benefits](https://janko.io/upgrading-from-selenium-to-cuprite/) 🙌🏻 

I decided we needed the new headless mode when I encountered a system test that was failing in old headless mode and passing in headful mode. We were able to use the new headless mode with current Cuprite by passing `browser_options: { "headless" => "new" }` to the driver, so not a blocker at all.

route:
@janko Very good article btw!

PR #382, opened 2023-07-23
Protect against calling `.quit` on the browser object twice.

postmodern:
Discovered that you can accidentally call `.quit` twice on the `browser` instance, and that causes the following exception:

```
/home/postmodern/test/ruby/bundler/vendor/bundle/ruby/3.2.0/gems/ferrum-0.13/lib/ferrum/browser.rb:240:in `quit': undefined method `close' for nil:NilClass (NoMethodError)

      @client.close
             ^^^^^^
	from test.rb:13:in `<main>'
```

### DIFF
--- a/lib/ferrum/browser.rb
+++ b/lib/ferrum/browser.rb
@@ -237,9 +237,11 @@ module Ferrum
     end
 
     def quit
-      @client.close
-      @process.stop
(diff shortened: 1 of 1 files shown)


route (approved):

route:
Thank you!

PR #399, opened 2023-09-13
fix: Chrome doesn't change resolution

route:


route:
@aaronjensen interesting but now I've got failing tests on my laptop, when this deviceScaleFactor is set to 0, though same Chrome version on github doesn't show broken tests. Same main branch was working for me for today before my experiments, but now it constantly failing for 117.

aaronjensen:
What are the failures? I can see if I can reproduce.

aaronjensen:
I see exactly one failing test, and the failure does seem related, but it may point to Chrome having changed their default scaling for headless browsers to 2x, which means that the test may be the thing that needs to be corrected. If the test produces different results on different machines, it could be that Chrome is somehow taking into account the device's scaling (I do have a high dpi monitor and my `window.devicePixelRatio` is 2).

```
  1) Ferrum::Page::Screenshot#screenshot supports screenshotting the whole of a page that goes outside the viewport
     Failure/Error: expect(ImageSize.new(f.read).size).to eq(browser.viewport_size)

       expected: [1024, 768]
            got: [2048, 1536]

       (compared using ==)
     Shared Example Group: "screenshot screen" called from ./spec/page/screenshot_spec.rb:158
     # ./spec/page/screenshot_spec.rb:16:in `block (4 levels) in <top (required)>'
     # ./spec/page/screenshot_spec.rb:15:in `open'
     # ./spec/page/screenshot_spec.rb:15:in `block (3 levels) in <top (required)>'
```

aaronjensen:
I propose that we allow deviceScaleFactor to be controlled the same way height and width is. Then the test can be set up to specify that. It should default to 0 so that it will respect the machine it is running on.

route:
@aaronjensen you should remove `fail-fast` in `.rspec` file and you can see more failing tests. What OS are you on?

aaronjensen:
macOS Ventura

As a point of feedback, it is maximally surprising to have rspec default to fail-fast. That's an end-user choice. I would have never known until I got incredibly frustrated playing test whack-a-mole.

aaronjensen:
All of the failing tests are exactly the same. It is respecting my device scaling now. As a result all screenshots are higher resolution. That is, they are more accurate to what my device should actually produce. I'd suggest either allowing device pixel ration to be controlled and updating those tests to control it, or updating the tests to test for a multiple of the expected size, rather than the exact size if, in fact, the CI server is passing these tests with Chrome 117

aaronjensen:
Alternatively, the tests can be corrected to expect pixel size when they expect pixel size, rather than the virtual device size with something like this:

```rb
def viewport_pixel_size
  evaluate <<~JS
    [window.innerWidth * window.devicePixelRatio,
     window.innerHeight * window.devicePixelRatio]
  JS
end
```

aaronjensen:
@route if you let me know what you prefer, I can work on a PR later, I think.

aaronjensen:
The change looks good, thanks!

PR #410, opened 2023-09-28
feat: Supports capture area screenshot

### LINKED ISSUE #389: Send Emulation commands to Page
Hey,

I am working on automatic rendering with Ferrum. I am interested in sending some additional Emulation commands.
For example, puppeteer has emulateViewport.

Can I contribute a few methods to Page or a Mixin?

erickgskovik:
Fixes #389

I made `selector` and `area` optional arguments. In the implementation, I think that the pull requests shows the true intention of the screenshot region.

If `full: true`, It tries to capture full screenshot with full size browser window and then viewport if the full size is not accurate.
If `full: false`, `:selector` is preferred if defined. Then `:area`. Otherwise, the default will be the viewport.

Applying default scale (1.0) will not affect the result anyway.

### DIFF
--- a/README.md
+++ b/README.md
@@ -359,7 +359,12 @@ Saves screenshot on a disk or returns it as base64.
   * :format `String` "jpeg" | "png"
   * :quality `Integer` 0-100 works for jpeg only
   * :full `Boolean` whether you need full page screenshot or a viewport
-  * :selector `String` css selector for given element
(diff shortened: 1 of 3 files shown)


sandstrom:
@route friendly ping 😄

route:
I'm back and going one by one from top to bottom ;)

route:
Thanks, nice job!

PR #414, opened 2023-10-19
Stop process if error during initialise

20TRIES:
We use Ferrum with Puma and found that if an error is raised during initialisation of the browser, then chrome processes do not get stopped until the garbage collection runs. 

Equally there is no way to stop these processes from the perspective of somebody using Ferrum because we never get a browser object back as the initialize method is raising the error.

From what i can tell, garbage collection doesn't run immediately and so these processes can live for longer then they need to.

By catching any errors raised between the process being started and returning from the browser initialiser then we can ensure that we stop any processes which were started.

### DIFF
--- a/lib/ferrum/browser.rb
+++ b/lib/ferrum/browser.rb
@@ -266,11 +266,21 @@ module Ferrum
 
     def start
       Utils::ElapsedTime.start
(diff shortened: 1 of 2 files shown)


route:
Thank you! Looks good!

PR #417, opened 2023-11-16
Fix: Do not expect response for initiatorType Ping requests

MC-Squared:
## Motivation
HTML supports adding a `ping` attribute to `<a>` tags. In such cases, the browser sents a POST request to the set URL(s) before navigating to the new URL (See [w3schools doc here](https://www.w3schools.com/TAGs/att_a_ping.asp) and [Mozilla docs on initiatorType](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceResourceTiming/initiatorType) which includes `ping` as an option).
Unfortunately, because these requests do not receive a server response Ferrum adds them to the list of `pending_connections`, meaning that subsequent calls to `wait_for_idle` will timeout waiting for these never-will-complete requests.

## Changes
* Add `Request#response_expected?` method to indicate if the request expects a response from the server (currently 'ping' is the only request I know of that would not get a response)
* Add `Exchange#response_expected?` that calls the request method. Note, if `request.nil?` we default to `true` to match the current behaviour
* Update `Exchange#finished?` so that any requests that do not expect responses are now also considered 'finished'

### DIFF
--- a/lib/ferrum/network/exchange.rb
+++ b/lib/ferrum/network/exchange.rb
@@ -119,6 +119,17 @@ module Ferrum
       end
 
       #
+      # Determines if the exchange expects a response
+      #
+      # @return [Boolean]
+      #
+      def response_expected?
+        return true if request.nil?
+
+        !!request.response_expected?
+      end
+
+      #
       # Returns request's URL.
       #
       # @return [String, nil]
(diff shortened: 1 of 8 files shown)


route:
Could you add a proper test with <a> tag?

route:
ping

MC-Squared:
Updated - I have guessed a bit on where best to locate the new specs so if they need to be moved/re-arranged just let me know

route:
Thanks! Merging, though I have some patch to apply on it after merging.

PR #425, opened 2023-12-13
introduce resize_windows option

decaffeinatedio:
Allow specification of whether the windows get automatically resized.


route:
This was a rudimentary artefact of Cuprite. I've moved it there, no need for this option.
Let me know if it solves the issue.

https://github.com/rubycdp/ferrum/pull/427

Issue #420, opened 2023-11-24
`wait_for_network_idle` failing due to pending requests on previous page


cbliard:
I experienced some random `Ferrum::TimeoutError` when calling `wait_for_idle`. For the same navigation, sometime it would fail, sometime it would pass.

After adding debug information output for `requestWillBeSent`, `responseReceived`, `loadingFinished` and `loadingFailed` events, it looks like some request can be sent while navigating to another page, and never receive a `responseReceived`, `loadingFinished`, or `loadingFailed` event.

For instance:
```
requestWillBeSent for requestId=349E42C73C6810F5DBFA790494E57AE0 loaderId=349E42C73C6810F5DBFA790494E57AE0 at 73687.356628
requestWillBeSent for requestId=429934.40 loaderId=8FCB3DE497DC83193F1D76075900D0D4 at 73687.362944
requestWillBeSent for requestId=429934.41 loaderId=8FCB3DE497DC83193F1D76075900D0D4 at 73687.452702
responseReceived for requestId=349E42C73C6810F5DBFA790494E57AE0 loaderId=349E42C73C6810F5DBFA790494E57AE0 at 73687.54549
```

The two requests `429934.40` and `429934.41` are in the `@traffic` array, but no `responseReceived`, `loadingFinished`, or `loadingFailed` event will be sent for them, so they will still be considered pending. They have a different `loaderId` because they have been loaded by the previous page doing some asynchronous XHR request for some reason.

When `wait_for_idle` is called, it will wait for all pending request to finish, even those from no-more displayed pages, making it fail with `Ferrum::TimeoutError`.

I modified the `#pending_connections` method a little to consider only pending connections for the current page. I looks somewhat like this:
```ruby
    def pending_connections
      frame_id = @traffic.first&.request&.frame_id
      current_page_loader_id = @traffic.select { |conn| conn.navigation_request?(frame_id) }.last.request.loader_id
      current_page_traffic = @traffic.filter { |exchange| exchange.request.loader_id == current_page_loader_id }
      current_page_traffic.count(&:pending?)
    end
```

This fixes my issue.

I am no expert in CDP. Maybe there are some other events that I missed. Is this approach correct?

Chrome version is 119.0.6045.105 on Linux amd64.

route:
@cbliard do you have an example of such a page in the public internet? I just want to check the CDP logs to better understand of what's going on

route:
merged https://github.com/rubycdp/ferrum/pull/426

PR #426, opened 2023-12-15
Handle pending connections losing context on frame navigation

francisbeaudoin:
# Details
Similar to what is being proposed in #420, this PR fixes intermittent `Ferrum::TimeoutError` as the driver awaits for requests to be completed.

# What
It was observed that in some scenarios the driver is not always receiving and associated `Network.loadingFailed` for requests that were in-flight as the main frame navigates. This may be happening with either main frame requests or sub-frame ones. 

I'm not sure if this is the right approach, but it appears to be working. Basically, it flags requests which are no longer relevant to the current page context as `unknown` which are then assumed to be `#finished`.

Relates to #420

### DIFF
--- a/lib/ferrum/network.rb
+++ b/lib/ferrum/network.rb
@@ -74,11 +74,11 @@ module Ferrum
     end
 
     def total_connections
-      @traffic.size
+      exchange_connections.count
     end
 
     def finished_connections
-      @traffic.count(&:finished?)
+      exchange_connections.count(&:finished?)
     end
 
     def pending_connections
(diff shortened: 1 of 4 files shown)


route:
@francisbeaudoin do you have an example of such a page in the public internet? I just want to check the CDP logs to better understand of what's going on. A test describing the issue would be more appropriate though.

route:
Hmmmm, nice thank you! Let me experiment a bit with that

route:
@francisbeaudoin yes sorry I takes me so long

route:
I guess we can move forward with some tweaks, I'll add them while rebasing

route:
Manually merged with some minor changes:
1. `total_connections` and `finished_connections` work the same way w/o breaking changes. They do return all requests, where some of them might be marked as unknown. The map/reduce code to filter out irrelevant exchanges might be added in users code base easily.
2. `loader_id` added to exchange
3. added a test case

route:
Thank you!

PR #429, opened 2023-12-27
chore: update ruby in workflows

erickgskovik:
I removed ruby 2.6 in tests since ruby 2.6 has been EOL for 1 year and 9 month ago.

Hope ruby 3.3 runs all smoothly.


route:
Awesome, thanks!

PR #434, opened 2024-01-06
feat: Flatten mode

route:


erickguan:
Nice, does `websocket-driver` multiplexes the same socket for different targets?

route:
@erickguan yes one socket is used to connect to the browser as well as targets.

Issue #439, opened 2024-01-25
Browserless 2.0 support


Fodoj:
Did anyone manage to make it work with Browserless 2.0? I am running everything in containers, meaning my rails app container connects to browsleress container. Even with the HEAD version, if I pass ws_url, it still stays localhost:4000 and tests are failing:

### Cuprite

```
Capybara.register_driver(:browserless) do |app|
  Capybara::Cuprite::Driver.new(
    app, 
    **{
      host: 'chrome', #'
      url: 'http://172.23.0.3:4000', 
      ws_url: 'ws://chrome:4000', 
      browser_options: { 
        headless: true,
        'no-sandbox': nil
      },
      process_timeout: 15,
      slowmo: 0.1,
      headless: "new",
      timeout: 20,
      window_size: [ 1920, 2048 ], 
    }
  )
end
```

### Ferrum

https://github.com/rubycdp/ferrum/blob/main/lib/ferrum/browser.rb#L242

```
    242: def start
    243:   Utils::ElapsedTime.start
    244:   binding.pry
    245:   @process = Process.new(options)
    246: 
    247:   begin
 => 248:     @process.start
    249:     @options.default_user_agent = @process.default_user_agent
    250: 
    251:     @client = Client.new(@process.ws_url, options)
    252:     @contexts = Contexts.new(@client)
    253:   rescue StandardError
    254:     @process.stop
    255:     raise
    256:   end
    257: end
```

Result:

```
[6] pry(#<Capybara::Cuprite::Browser>)> options.ws_url
=> "ws://chrome:4000"
[7] pry(#<Capybara::Cuprite::Browser>)> @process.ws_url
=> #<Addressable::URI:0x7f1c URI:ws://0.0.0.0:4000>
```

Because of this my tests won't even start, as they quickly fail during initialization phase.

Fodoj:
I have this setup already working. The article was written in 2020, Browserless 2.0 was released in December 2023. Browserless 1.x works without any issues, but new one works as describe in the issue above.

Fodoj:
See also: https://github.com/browserless/browserless/blob/main/MIGRATION-2.0.md#browserless-20-migration-guide

route:
Got it, I'll check then.

Aubermean:
Did you ever manage to get it working or find out what the issue was?

route:
It's not my top priority but it works, so I don't know what is not working:

```ruby
require "ferrum"

[...]

PR #447, opened 2024-02-22
Better handling of ws_url, and avoid stripping query params from url

Aubermean:
Hi,

This fixes some compatibility issues with [browserless](https://github.com/browserless/browserless) v2 and possibly others.

An issue with the `parse_json_version` was that it would turn the ws scheme into http, strip query params and request /json/version. This is the first problem, stripping query params which often include an auth token and other configuration.

The second concern was that the `initializer` would override the ws_url based on what that endpoint would show, in the case of browserless that would be 0.0.0.0:3000. If the ws_url has been set, ferrum should not try to override that.

This is my first OSS PR so I am sorry if I did some thing wrong.

#439

### DIFF
--- a/lib/ferrum/browser/process.rb
+++ b/lib/ferrum/browser/process.rb
@@ -64,11 +64,9 @@ module Ferrum
 
         if options.ws_url
           response = parse_json_version(options.ws_url)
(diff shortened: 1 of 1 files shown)


route:
Tests are failing

route:
Merged with other fixes manually.

PR #453, opened 2024-03-19
fix: Support all Ruby 2.7 for now

route:
Ruby syntax `...` was added not for all 2.7 versions. Let's fallback for now to old syntax.

### DIFF
--- a/lib/ferrum/client.rb
+++ b/lib/ferrum/client.rb
@@ -34,8 +34,8 @@ module Ferrum
       @client.respond_to?(name, include_private)
     end
 
-    def method_missing(name, ...)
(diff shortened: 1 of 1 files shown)


erickguan:
Hey! Do you consider bumping to newer ruby versions? Ruby 3.1 is end of life next week or so. Other projects, e.g. rails, start to require 3.1.

This would allow using newer Ruby syntax. Additionally, you could build a new release and then bump the required version.

route:
Newer ruby syntax is not a requirement now for ferrum. There are still projects that still use 2.7.
I will definitely in a couple of releases do that.

PR #455, opened 2024-03-23
Allow subscriptions to be removed

rzane:
This is admittedly not the most elegant API, but my goal is to start a discussion about how this should look.

Here's how you'd subscribe and unsubscribe:
```ruby
subscription = page.on("Some.event") do |params, _index, _total|
  # etc
end

page.off("Some.event", subscription)
```

Here are some important considerations:
* The `#on` method previously returned `true`. Now, it returns the index of the block within `@on[event]`. This could be considered a breaking change. Rather than using `#on` and `#off`, we could introduce a new method called `#subscribe`.
* Rather than returning a value that identifies the specific event handler, we could return an object that has an `#unsubscribe` method.

### DIFF
--- a/lib/ferrum/client.rb
+++ b/lib/ferrum/client.rb
@@ -26,6 +26,10 @@ module Ferrum
       @client.on(event_name(event), &block)
     end
 
+    def off(event, id)
+      @client.off(event_name(event), id)
+    end
+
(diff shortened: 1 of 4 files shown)


route:
Thank you! Looking solid now!

PR #464, opened 2024-06-18
Fixed a broken JSON in log output

willnet:
Since v0.15, a JSON format in the log output has been broken. When the original log is in the following format:

```
"{\"id\":27,\"result\":{\"data\":\"(snipped)\"},\"sessionId\":\"E2C753A9154794C390311B8647590A03\"}"
```

It gets converted to a format without a closing } like this:

```
"    ◀ 10.680710212999657 {\"id\":27,\"result\":\"Set FERRUM_LOGGING_SCREENSHOTS=true to see screenshots in Base64\"\n"
```

Upon investigation, I found that the string replacement performed when SKIP_LOGGING_SCREENSHOTS is enabled was unintentionally breaking the JSON format. Therefore, I modified the regular expression used for the replacement.

### DIFF
--- a/lib/ferrum/client/web_socket.rb
+++ b/lib/ferrum/client/web_socket.rb
@@ -63,7 +63,7 @@ module Ferrum
         output = event.data
         if SKIP_LOGGING_SCREENSHOTS && @screenshot_commands[data["id"]]
           @screenshot_commands.delete(data["id"])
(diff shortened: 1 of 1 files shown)


botandrose:
+1 from me. Came here to submit more or less the same PR, just making the regex non-greedy. `/{"data":"(.*?)"}/`.

route:
We need to address linter issues

Issue #47, opened 2020-02-12
Allow Ferrum to use default (non incognito) context


Pugio:
I'd like my browser state/session to be preserved between uses of my script.

Currently, Ferrum places all commands inside a BrowserContext, which acts like an Incognito session and doesn't load stored cookies.

I would like to use the browser's default session (which isn't inside a context), and take advantage of the `--user-data-dir` flag to be able to persist state across sessions. This would require modifying how Ferrum uses and creates contexts to allow for non-browserContext requests. I am happy to do this, and submit a PR, but would like some author feedback (if possible) about whether such an option is in line with the project's vision before I start working on code for this.

route:
Absolutely, go ahead! I just didn’t need it thus it was easy not to implement rather than implement. I think it will be as easy as provide a sit and just don’t remove anything from it when quitting.

route:
I think the default context that Chrome creates has a lot of limitations like inability to create pages inside it and so on. If you come up with something useful please open a PR but due to inactivity I'm going to close this issue.

yann120:
Hi,
Is there an other way to persist state between sessions like we can do with Selenium ?
I came to Ferrum to be able to catch network requests, but I need also to persist the session.
Exemple, i'm logged in on a website, I want to be still logged when re-launching the script
Thanks 🙏🏻

route:
@yann120 http is stateless protocol. The state is determined by headers sent from/to web server. If you want to login on the website, close browser and then open it and be logged in, you need to store your cookies somewhere, and next time before telling a browser to visit url, load those cookies.

yann120:
> @yann120 http is stateless protocol. The state is determined by headers sent from/to web server. If you want to login on the website, close browser and then open it and be logged in, you need to store your cookies somewhere, and next time before telling a browser to visit url, load those cookies.
> 
> You open the page and login:
> 
> ```ruby
> browser = Ferrum::Browser.new(headless: false)
> page = browser.create_page
> page.go "https://auth.udacity.com/sign-in?next=https%3A%2F%2Flearn.udacity.com%2F"
> page.network.wait_for_idle
> 
> page.at_xpath("//*[@id='email']").focus
> page.keyboard.type(email)
> page.at_xpath("//*[@id='revealable-password']").focus
> page.keyboard.type(password)
> button = page.at_xpath("//button[@type='button'][./span[.='Sign in']]")
> button.focus.click
> 
> page.network.wait_for_idle
> 
[...]

PR #471, opened 2024-07-04
Allow using default browser context

ibrahima:
This PR adds an option for Ferrum to use the default browser context instead of a created BrowserContext that doesn't have access to the persisted browser state.

I know that historically for testing purposes it's been recommended here to use clean browser contexts for reproducibility, but I think this is a reasonable option for some use cases (that may not be related to testing). Other similar libraries provide such options. For example, Playwright lets you do [launchPersistentContext](https://playwright.dev/docs/api/class-browsertype#browser-type-launch-persistent-context) to launch a browser with a persistent context. Puppeteer seems to default to using the default browser context and [only creates new BrowserContexts if you manually create them](https://pptr.dev/guides/browser-management#browser-contexts). I don't have experience with these libraries, but I did a lot of digging through their source to see how they handled Target creation and BrowserContext creation.

Fixes https://github.com/rubycdp/ferrum/issues/47

Re: this comment from that thread:

> I think the default context that Chrome creates has a lot of limitations like inability to create pages inside it and so on.

I seem to be able to create pages/targets within the default context fine. However, I haven't explored much so there could very well be bugs associated with this change. All I know is that it works for my use case and it would be very useful to have this in the upstream gem. I'm happy to discuss whether this change makes sense or not and I don't mind if the answer is that it doesn't. But at least in my case, the save/load cookies feature from 99cfa84c56e55bb373da48935d08d5a13df1ad27 doesn't solve my use case because the persisted browser state that I'm trying to restore is coming from a browser hosting service (https://www.browserbase.com/), so the only way I can access the state is through the browser's default context.

Thank you for your consideration!

### DIFF
--- a/lib/ferrum/browser.rb
+++ b/lib/ferrum/browser.rb
@@ -125,6 +125,9 @@ module Ferrum
     # @option options [Hash] :env
     #   Environment variables you'd like to pass through to the process.
     #
+    # @option options [Boolean] :use_default_context
+    #   When true, allows using the default browser context that has access to the browser's persistent state.
+    #
     def initialize(options = nil)
       @options = Options.new(options)
       @client = @process = @contexts = nil
--- a/lib/ferrum/contexts.rb
+++ b/lib/ferrum/contexts.rb
@@ -11,6 +11,7 @@ module Ferrum
     def initialize(client)
       @contexts = Concurrent::Map.new
       @client = client
+      @default_context = create_default_context
       subscribe
       auto_attach
       discover
@@ -18,6 +19,20 @@ module Ferrum
 
     def default_context
       @default_context ||= create
+    end
+
+    def create_default_context
+      default_context_id = compute_default_context_id
+      # Targets created in this context will not be created with a browserContextId
+      @contexts[default_context_id] = ::Ferrum::Context.new(@client, self, nil)
+    end
+
+    # Compute the default context ID by looking for contexts not returned by Target.getBrowserContexts
+    def compute_default_context_id
+      created_contexts = Set.new(@client.command("Target.getBrowserContexts")["browserContextIds"])
+      targets = @client.command("Target.getTargets")["targetInfos"]
+      all_contexts = Set.new(targets.map { |target| target["browserContextId"] })
+      (all_contexts - created_contexts).first
     end
 
     def each(&block)
--- a/lib/ferrum/contexts.rb
+++ b/lib/ferrum/contexts.rb
@@ -24,7 +24,7 @@ module Ferrum
     def create_default_context
       default_context_id = compute_default_context_id
       # Targets created in this context will not be created with a browserContextId
-      @contexts[default_context_id] = ::Ferrum::Context.new(@client, self, nil)
+      @contexts[default_context_id] = Context.new(@client, self, nil)
     end
 
     # Compute the default context ID by looking for contexts not returned by Target.getBrowserContexts
--- a/lib/ferrum/contexts.rb
+++ b/lib/ferrum/contexts.rb
@@ -11,7 +11,7 @@ module Ferrum
     def initialize(client)
       @contexts = Concurrent::Map.new
       @client = client
-      @default_context = create_default_context
+      @default_context = create_default_context if @client.options.use_default_context
       subscribe
       auto_attach
       discover
--- a/lib/ferrum/browser/options.rb
+++ b/lib/ferrum/browser/options.rb
@@ -15,7 +15,7 @@ module Ferrum
                   :js_errors, :base_url, :slowmo, :pending_connection_errors,
                   :url, :ws_url, :env, :process_timeout, :browser_name, :browser_path,
                   :save_path, :proxy, :port, :host, :headless, :browser_options,
-                  :ignore_default_browser_options, :xvfb, :flatten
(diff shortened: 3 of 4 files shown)


lib/ferrum/browser.rb
```diff
@@ -125,6 +125,9 @@ class Browser
     # @option options [Hash] :env
     #   Environment variables you'd like to pass through to the process.
     #
+    # @option options [Boolean] :use_default_context
```
ibrahima, line 128: I don't know what the best name for this is... it's kinda confusing since `Contexts#default_context` is already a thing and the current behavior of that method is to create a new BrowserContext. But I felt like "default" should be the right thing because the effective change is to not pass a browserContextId when [creating a target](https://chromedevtools.github.io/devtools-protocol/tot/Target/#method-createTarget).

Perhaps a less confusing name would be `use_persistent_context`? Similar to how Playwright has `launchPersistentContext`.

lib/ferrum/contexts.rb
```diff
@@ -11,6 +11,7 @@ class Contexts
     def initialize(client)
       @contexts = Concurrent::Map.new
       @client = client
+      @default_context = create_default_context if @client.options.use_default_context
```
ibrahima, line 14: One side effect of setting this in the constructor is that any existing tabs/targets in the browser's default context will be available as targets (e.g. `browser.default_context.targets` will return a nonzero number of targets). This felt right to me, because then it handles use cases like https://github.com/rubycdp/ferrum/issues/320 as well. Similar to the fork mentioned in https://github.com/rubycdp/ferrum/issues/320#issuecomment-1368368807 but in my opinion cleaner because it just gives you access to all the existing tabs to do with what you will, rather than somewhat arbitrarily picking the default (latest?) tab to be the default target.

lib/ferrum/contexts.rb
```diff
@@ -20,6 +21,20 @@ def default_context
       @default_context ||= create
     end
 
+    def create_default_context
+      default_context_id = compute_default_context_id
+      # Targets created in this context will not be created with a browserContextId
+      @contexts[default_context_id] = ::Ferrum::Context.new(@client, self, nil)
+    end
+
+    # Compute the default context ID by looking for contexts not returned by Target.getBrowserContexts
```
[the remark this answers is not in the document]
ibrahima: Hm this method was also suggested (independently) by one of the Chromium engineers, so it might be fine. I am slightly nervous about the case when no tabs exist in the default context which presumably could happen, but I think generally people using browser automation should be able to ensure that the correct prerequisites are met.

One other approach might be to have the `Contexts` class (optionally) enumerate all existing contexts at startup and create entries for them, and then any targets that aren't part of the discovered contexts are probably from the default context. This could get out of sync potentially, but I'm not sure if there is really any problem with a Target (in Ruby) being assigned to the incorrect Context object from the Contexts map, since the Target seems to have sufficient information to operate on its own and doesn't really need info from the Context. I think the main purpose of that association is that when `Context#dispose` is called, it closes the connections to the `Targets` in that `Context`, so that's probably not going to be an issue in practice because if Ferrum is not aware of that `Context` you're never going to call `#dispose` on it.

route:
@ibrahima I'm on vacation currently and will take a look on Monday or sooner if I have time

yann120:
Hey @ibrahima, thanks for this PR.
This is exactly the feature I am missing to moving to Ferrum!
I need to stay logged in on a website, and don't want to log in on the website every 5 minutes

@route have you found time to review this PR ? 🙈
Thanks a lot!

matti:
@route how about reviewing this?

sebyx07:
@yann120 also reusing the chrome data dir will help you

ryanstout:
Hey @ibrahima. Thanks for working on this! I was trying to add the same functionality.

I'm seeing where if I run the following:

```
require "ferrum"

browser = Ferrum::Browser.new(
  headless: false,
  use_default_context: true,
)

browser.go_to("https://google.com")
```

I get:

```
/Users/ryanstout/Sites/arsenal/instagrab_2024/ferrum/lib/ferrum/context.rb:52:in `create_target': Ferrum::NoSuchTargetError (Ferrum::NoSuchTargetError)
        from /Users/ryanstout/Sites/arsenal/instagrab_2024/ferrum/lib/ferrum/context.rb:20:in `default_target'
        from /Users/ryanstout/Sites/arsenal/instagrab_2024/ferrum/lib/ferrum/context.rb:24:in `page'
        from /Users/ryanstout/.asdf/installs/ruby/3.2.2/lib/ruby/3.2.0/forwardable.rb:240:in `page'
        from /Users/ryanstout/.asdf/installs/ruby/3.2.2/lib/ruby/3.2.0/forwardable.rb:234:in `go_to'
        from test.rb:15:in `<main>'
        ```
        
I'm digging in on it, but I wasn't sure if you had seen this. Thanks!

yann120:
> @yann120 also reusing the chrome data dir will help you

Thanks, I ended up saving the cookies, and reload them, and it works. Maybe the easiest solution :)

route:
I don't think it works as expected, the issue is not getting contexts and targets but running a browser with --no-startup-window flag which is not addressed here. Superseded with https://github.com/rubycdp/ferrum/pull/532

PR #476, opened 2024-07-24
Add base64 to runtime dependencies for Ruby 3.4 compatibility

tisba:
See https://github.com/ruby/ruby/blob/master/doc/NEWS/NEWS-3.3.0.md#stdlib-updates

With Ruby 3.4 on forward, `base64` will no longer be available as part of part of the default gems. It needs to be an explicit dependency going forward.

### DIFF
--- a/ferrum.gemspec
+++ b/ferrum.gemspec
@@ -26,6 +26,7 @@ Gem::Specification.new do |s|
   s.required_ruby_version = ">= 2.7.0"
 
   s.add_runtime_dependency "addressable",      "~> 2.5"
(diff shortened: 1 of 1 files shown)


tisba:
Dear maintainers, let me know if you want me to fix the linting errors too. I kept the change in this PR to a minimum for now.

route:
Let's fix linter and I'll merge!

tisba:
Linters are happy. The now failing `Ferrum::ProcessTimeoutError` for Ruby 2.7 looks like an intermittent error.

route:
Thanks!

PR #487, opened 2024-08-27
Improve thread scheduling

sebyx07:
use Thread.pass where ever possible, because ferrum scrapers usually are put in sidekiq workers and it's better to yield to other threads that actually do work instead of waiting
https://github.com/sidekiq/sidekiq/discussions/5039

and `wait_for_idle(..duration:)` the duration param is not really set by someone duration:, only the timeout value is usually set

### DIFF
--- a/lib/ferrum/browser/process.rb
+++ b/lib/ferrum/browser/process.rb
@@ -16,7 +16,6 @@ module Ferrum
   class Browser
     class Process
       KILL_TIMEOUT = 2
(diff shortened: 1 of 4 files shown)


sebyx07:
the specs fail locally on master. idk if u can check them out?

sebyx07:
https://github.com/rubycdp/ferrum/pull/488

route:
I'd say maybe, but in general it's bad idea to run crawlers in sidekiq

PR #507, opened 2025-01-02
Implement screencasting 

singhprd:
This only saves images, it's not true video/screencasting.

They should be very helpful for viewing tests however.


lib/ferrum/page.rb
```diff
@@ -64,17 +67,9 @@ class Page
...
-    #
-    # @return [Downloads]
-    attr_reader :downloads
-
-    def initialize(client, context_id:, target_id:, proxy: nil)
-      @client = client
-      @context_id = context_id
-      @target_id = target_id
-      @options = client.options
+    attr_reader :screencaster
```
singhprd, line 70:
```suggestion
    attr_reader :screencaster
    
    # Downloads object.
    #
    # @return [Downloads]
    attr_reader :downloads

    def initialize(client, context_id:, target_id:, proxy: nil)
      @client = client
      @context_id = context_id
      @target_id = target_id
      @options = client.options
```

lib/ferrum/page.rb
```diff
@@ -467,6 +484,11 @@ def prepare_page
 
       inject_extensions
 
+      width, height = @browser.window_size
+      resize(width: width, height: height)
+
+      start_screencast
```
singhprd, line 490:
```suggestion
```

lib/ferrum/page.rb
```diff
@@ -69,12 +72,24 @@ class Page
     # @return [Downloads]
     attr_reader :downloads
 
+    attr_reader :screencaster
+    
+    # Downloads object.
+    #
+    # @return [Downloads]
+    attr_reader :downloads
+
```
singhprd, line 81:
```suggestion
```

lib/ferrum/screencaster.rb
```diff
@@ -0,0 +1,26 @@
...
+      def initialize(page)
+        @page = page
+      end
+
+      def add_frame(params)
+        warn 'frame'
+        @page.command("Page.screencastFrameAck", sessionId: params["sessionId"])
+
+        ts = (Time.now.to_f * 1000).to_i
+        File.binwrite("img_#{ts}.jpeg", Base64.decode64(params["data"]))
```
singhprd, line 14: Could maybe put this in a background thread to stop from blocking and get more frames per second.

PR #522, opened 2025-02-19
Fix: Safe call body method (DOM) when page is empty

Mihoid:
In some cases method _body_ (DOM) can raise an exception. It happens when page is empty and there is no _html_
e.g. 
```
=> browser = ::Ferrum::Browser.new
=> browser.goto('https://somesitewithemptypage.com') # <—— here we receive empty page
=> browser.body 
=> 
Ferrum::JavaScriptError: TypeError: Cannot read properties of null (reading 'outerHTML') |  
```
This PR will add safe call method _body_ for reducing errors

### DIFF
--- a/spec/network_spec.rb
+++ b/spec/network_spec.rb
@@ -517,6 +517,6 @@ describe Ferrum::Network do
       %r{Request to http://.*/with_js failed \(net::ERR_INTERNET_DISCONNECTED\)}
     )
 
-    expect(page.at_css("body").text).to match("No internet")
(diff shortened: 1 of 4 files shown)


spec/network_spec.rb
```diff
@@ -517,6 +517,6 @@
       %r{Request to http://.*/with_js failed \(net::ERR_INTERNET_DISCONNECTED\)}
     )
 
-    expect(page.at_css("body").text).to match("No internet")
```
Mihoid, line 520: this changes little bit confusing me 🤔

PR #523, opened 2025-02-23
feat: Add `Frame#parent` method

myabc:
Facilitates easier traversal of nested `iframe`s.


myabc:
Not sure if test failures on 3.0 and 3.2 relate.

route:
Thanks!

PR #524, opened 2025-02-23
feat: Add `Frame#frame_element` method

myabc:
⚠️ ~~**Includes commit from #523** - please review/merge that PR first.~~

Faciliates easier retrieval of the `iframe` element that embeds the `window` (represented as a `Frame` object by Ferrum).

This is extracted from my Cuprite PR here: https://github.com/rubycdp/cuprite/pull/291 _(that PR doesn't currently depend on this PR being merged, but the Cuprite PR could easily be adapted to call the newly-added Ferum `Frame#node` method)_.

### DIFF
--- a/lib/ferrum/frame/dom.rb
+++ b/lib/ferrum/frame/dom.rb
@@ -96,6 +96,22 @@ module Ferrum
       end
 
       #
+      # Returns the element in which the window is embedded.
+      #
+      # @return [Node, nil]
+      #   The element in which the window is embedded.
+      #
+      # @example
+      #   browser.go_to("https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe")
+      #   frame = browser.frames.last
+      #   frame.frame_element # => [Node]
+      #   frame.parent.parent.parent.frame_element # => nil
+      #
+      def frame_element
+        evaluate("window.frameElement")
+      end
+
+      #
       # Finds nodes by using a XPath selector.
       #
       # @param [String] selector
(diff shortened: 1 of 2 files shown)


lib/ferrum/frame/dom.rb
```diff
@@ -95,6 +95,22 @@ def body
...
+      # @return [Node, nil]
+      #   The node in which the window is embedded.
+      #
+      # @example
+      #   browser.go_to("https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe")
+      #   frame = browser.frames.last
+      #   frame.node # => [Node]
+      #   frame.parent.parent.parent.node # => nil
+      #
+      def node
```
route, line 110: I think the naming `node` is very confusing. What if we call this method on a node inside iframe? What node can be returned for a node? It should resemble js naming `frame_element` or `frame_node` or even both I think

myabc: @route thanks for the feedback.

My original need for this PR was to try to get the `Ferrum::Node` for a given `Ferrum::Frame` - hence I chose to call the method `#node` to match the corresponding return type (I also considered `#to_node`, `#as_node`, `#to_element`).

However I can agree that the naming could lead to confusion. Mirroring Web APIs in Ferrum isn't a bad idea either - and will likely aid Ferrum's API discoverability.

I've gone ahead and renamed the new method `#frame_element`. I haven't added an alias for `#frame_node` as I'm not 100% sure offering a choice of methods makes for a better overall API - let me know if want me to make that change though.

route: sounds good!

PR #527, opened 2025-03-21
Enable mobile emulation

### LINKED ISSUE #94: Expose API to enable mobile emulation mode
From what I see in the library there's only one call to `setDeviceMetricsOverride` that sets `mobile: false`. 

It would be great if there was an exposed API to enable mobile emulation and manage width and height of the viewport dynamically.

duncan-bayne:
There is a request on Cuprite to support mobile emulation: https://github.com/rubycdp/cuprite/issues/216

A prerequisite for this is for Ferrum to support `mobile: true` in `Page#resize`.  This PR adds that support.

Note that passing `mobile: true` also sets the screen resolution to that of an iPhone 14; see [this comment](https://github.com/rubycdp/ferrum/pull/527#pullrequestreview-2708886758).  Perhaps there's a nicer way of doing this?

Merging this should let us close #94.

### DIFF
--- a/lib/ferrum/page.rb
+++ b/lib/ferrum/page.rb
@@ -158,9 +158,22 @@ module Ferrum
         deviceScaleFactor: scale_factor,
         mobile: mobile
       )
+
+      options = if mobile
+                  {
+                    enabled: true,
+                    maxTouchPoints: 1
+                  }
+                else
+                  {
+                    enabled: false
+                  }
+                end
+
+      command("Emulation.setTouchEmulationEnabled", **options)
     end
 
-    def resize(width: nil, height: nil, fullscreen: false)
+    def resize(width: nil, height: nil, fullscreen: false, mobile: false)
       if fullscreen
         width, height = document_size
         self.window_bounds = { window_state: "fullscreen" }
@@ -169,7 +182,7 @@ module Ferrum
         self.window_bounds = { width: width, height: height }
       end
 
-      set_viewport(width: width, height: height)
+      set_viewport(width: width, height: height, mobile: mobile)
     end
 
     #
--- a/lib/ferrum/page.rb
+++ b/lib/ferrum/page.rb
@@ -137,7 +137,11 @@ module Ferrum
     end
 
     #
-    # Overrides device screen dimensions and emulates viewport according to parameters
+    # Overrides device screen dimensions and emulates viewport according to parameters.
+    #
+    # Note that passing mobile: true will cause set_viewport to ignore the passed
+    # height and width values, and instead use 390 x 844, which is the viewport size
+    # of an iPhone 14.
     #
     # Read more [here](https://chromedevtools.github.io/devtools-protocol/tot/Emulation/#method-setDeviceMetricsOverride).
     #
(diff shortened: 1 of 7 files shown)


lib/ferrum/page.rb
```diff
@@ -149,17 +153,39 @@ def close_connection
...
-        mobile: mobile
-      )
+      if mobile
+        command(
+          "Emulation.setTouchEmulationEnabled",
+          enabled: true,
+          maxTouchPoints: 1
+        )
+
+        command(
```
duncan-bayne, line 163: I'm uncertain about this.  Ideally the caller would always specify a mobile height, width, and scale factor.  But in principle that doesn't always happen, so I decided to override them here.  

Perhaps we should raise an exception if a height and width other than `0` are passed along with `mobile: true`?

spec/page_spec.rb
```diff
@@ -235,4 +235,55 @@
...
+
+  describe "#resize" do
+    def body_size
+      {
+        height: page.evaluate("document.body.clientHeight"),
+        width: page.evaluate("document.body.clientWidth")
+      }
+    end
+
+    def is_mobile?
```
duncan-bayne, line 247: This doesn't _really_ test whether mobile emulation is enabled, but I couldn't find a good way of doing that directly via JavaScript evaluation.  So we use touch emulation as a proxy.

Gemfile
```diff
@@ -4,6 +4,7 @@ source "https://rubygems.org"
 
 gem "byebug", "~> 11.0", platforms: %i[mri mingw x64_mingw]
 gem "chunky_png", "~> 1.3"
+gem "debug", "~> 1.10", require: false, platform: :mri
```
route, line 7: Please remove unnecessary changes not related to `Enable mobile emulation`

README.md
```diff
@@ -194,7 +194,7 @@ Ferrum::Browser.new(options)
   * `:proxy` (Hash) - Specify proxy settings, [read more](https://github.com/rubycdp/ferrum#proxy)
   * `:save_path` (String) - Path to save attachments with [Content-Disposition](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition) header.
   * `:env` (Hash) - Environment variables you'd like to pass through to the process
-
+  * `:mobile` (Boolean) - Specify whether to enable mobile emulation and touch UI.
```
route, line 197: I don't think we need flag for browser instantiation.

lib/ferrum/page.rb
route: Why setting it hard in the first place?

README.md
duncan-bayne: My intent was to expose this via driver registration, as per Selenium.  

I have a draft PR open for Cuprite that uses this; see e.g.:

https://github.com/rubycdp/cuprite/pull/292/files#diff-89eebfcbc0f14b6d989517837ca1e94fce4e2ce9a03233641cd936f2b8d2ed94

```
Capybara.register_driver(:cuprite_mobile) do |app|
  options = { mobile: true }
  options.merge!(inspector: true) if ENV["INSPECTOR"]
  options.merge!(logger: StringIO.new) if ENV["CI"]
  options.merge!(headless: false) if ENV["HEADLESS"] == "false"
  Capybara::Cuprite::Driver.new(app, options)
end
```

lib/ferrum/page.rb
duncan-bayne: It isn't, but it's also surprising - if we didn't hard-code the resolution here, users would have to know:

1. That they _need_ to set the resolution at the same time as enabling mobile, and
2. What the right resolution is for mobile devices.

Hence my suggestion that perhaps I should make this raise an exception if the user also passes height or width?  That way it should be really clear.

duncan-bayne: Have made it raise an exception now, so the behaviour is:

```
context "when a width is passed" do
  it "raises an error" do
    expect do
      page.resize(width: 100, height: 0, mobile: true)
    end.to raise_error("specifying `mobile: true` causes automatic resizing to mobile dimensions")
    end
  end
```

duncan-bayne: After reviewing with @DanielHeath, will *not* implement hard-coded resizing in Ferrum, but instead provide common mobile devices in Cuprite to choose from.

duncan-bayne:
Moving this back to draft, for now, while I clean up following a move of the mobile specific stuff into Cuprite.

duncan-bayne:
Okay I think this is good for review again :)

Note the new, improved, and not-hard-coded API.  In Cuprite you'd do:

```
Capybara.register_driver :custom_cuprite_mobile do |app|
  mobile_cuprite_options = cuprite_options.merge(Capybara::Cuprite::Devices::IPHONE_14)
  Capybara::Cuprite::Driver.new(app, mobile_cuprite_options)
end
```

duncan-bayne:
Apologies for the delay, but I've cleaned up this PR (now passes tests and linting), and it should be ready for review.

duncan-bayne:
@route My plan (assuming you're happy with this PR!) is to merge https://github.com/rubycdp/cuprite/pull/292 once this gets merged.

route:
It's going to be superseded by https://github.com/rubycdp/ferrum/pull/618 if we go down this path

PR #536, opened 2025-05-14
Fix for nextwork_spec.rb:512 "#offline_mode"

CodeMeister:
For me the message contains "No Internet", but the test expects "No internet".

The PR downcases the text before matching.


route:
That's good! Tests pass

PR #539, opened 2025-06-10
Fix Target.targetCreated handler for existing targets

markiz:
Came up in our large system spec suite, `Client` does not respond to `session_id`, so I'm assuming `target#session_id` should be right here. Haven't seen any errors related to that since.


route:
I guess you are running specs with ferrum `flatten: false`, any chance to switch to true?

markiz:
@route 

> I guess you are running specs with ferrum flatten: false

yeah. I'm afraid I don't remember the exact reason why we had to switch to flatten: false, but it probably had to do with stability. 

> I guess you are running specs with ferrum flatten: false, any chance to switch to true?

Yeah, checking it out. I'm seeing some performance regressions with flatten (5-6 minutes runs vs 3-4 minutes; the normal deviation for this particular suite is ~1 minute though), and also some intermittent `Browser is dead or given window is closed`. 

Is that going to be the new default or something in v1?

route:
Interesting, by any chance any reproducible test out of your project, at least for the error, perfomance regressions would be hard to debug, but very interesting.

markiz:
Ugh, I'm afraid that would be really difficult to isolate, it's a pretty large and tangled suite. I'll try to make some sense of it when I have some free time (that could take a while though)

route:
@markiz you could run test suite with `FERRUM_DEBUG=true rspec` and send me failing examples to email. I could at least fix that for your test suite, that might resolve performance thing as it's 30s wait for a dead browser. Anyways if you can send me something to dig for errors, and performance can be fixed as a consequence.

PR #543, opened 2025-08-22
CI: Tell dependabot to update GH Actions versions

olleolleolle:
This PR adds a configuration to tell Dependabot to update the versions used for some well-known GitHub Actions, such as `checkout`, in order to avoid deprecation warnings when those get too old. [Documentation](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot)

In addition,

- fix the permissions-setting step to point to the correct file
- avoid Gemfile-related warnings about gemified standard gems that

Not done here:

- make linting pass

Current test failing with

```
  1) Ferrum::Browser::Binary#find finds one binary
     Failure/Error: raise ProcessTimeoutError.new(timeout, output)
     Ferrum::ProcessTimeoutError:
       Browser did not produce websocket url within 25 seconds, try to increase `:process_timeout`. See https://github.com/rubycdp/ferrum#customization
     # ./lib/ferrum/browser/process.rb:170:in 'Ferrum::Browser::Process#parse_ws_url'
     # ./lib/ferrum/browser/process.rb:104:in 'Ferrum::Browser::Process#start'
     # ./lib/ferrum/browser.rb:277:in 'Ferrum::Browser#start'
     # ./lib/ferrum/browser.rb:136:in 'Ferrum::Browser#initialize'
     # ./spec/spec_helper.rb:48:in 'Class#new'
     # ./spec/spec_helper.rb:48:in 'block (2 levels) in <top (required)>'
```

Perhaps this is an issue with "the browser is now called something else". I'll see if I can get a hold of this.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -18,3 +18,8 @@ gem "sinatra", "~> 3.2"
 gem "yard", "~> 0.9", require: false
 
 gemspec
+
+if RUBY_VERSION > "3.4"
+  gem "logger" # "warning: logger was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0."
+  gem "ostruct" # "warning: ostruct was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0."
+end


route:
I think most of it already handled and tests are passing but please reopen if something it critical/forgotten or can be improved.

PR #559, opened 2026-01-28
Fix wss urls and session_id lost

neo87cs:
The current implementation of Ferrum is missing two important points:

1. the `ws_url` is missing support to `wss://...`, which is a SSL implementation;
2. cause of subscribers in `contexts.rb` being async, the `context.rb#add_target` was loosing the `session_id` property


route:
can you rebase on main? tests are fixed now.

PR #574, opened 2026-03-23
Prevent initialization failure

zz0ya:
Hi,
Fix for `can't modify frozen Hash` issue when initializing with `incognito: false` option.


route:
Thanks!

PR #582, opened 2026-04-28
fix: Browser#create_page ensure block masks original error with NoMethodError

collin-pereira:
## Problem

When `Browser#create_page` is called with `new_context: true` and a block, an exception raised by `contexts.create(...)` (for example `Ferrum::DeadBrowserError` when the browser dies before the context is created) is masked by a `NoMethodError` from the `ensure` block.

The local `context` is `nil` at that point because the assignment never completed. The `ensure` block then calls `context.dispose` unconditionally and raises `NoMethodError: undefined method 'dispose' for nil`, which propagates in place of the original error. Callers who rescue `Ferrum::Error` (the gem's documented contract) won't catch the masked `NoMethodError`, so legitimate errors like `DeadBrowserError` slip past their handlers.

## Fix

The adjacent cleanup of `page` already uses safe navigation (`page&.close`) for the same reason; this just makes `context.dispose` symmetric.

A regression spec stubs `browser.contexts.create` to raise `Ferrum::DeadBrowserError` and asserts the original error propagates through the block form. Without the fix it fails with `expected Ferrum::DeadBrowserError ... got NoMethodError: undefined method 'dispose' for nil`.

### DIFF
--- a/lib/ferrum/browser.rb
+++ b/lib/ferrum/browser.rb
@@ -172,7 +172,7 @@ module Ferrum
     ensure
       if block_given?
         page&.close
-        context.dispose if new_context
+        context&.dispose if new_context
       end
     end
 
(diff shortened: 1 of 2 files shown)


route:
Looks legit, thank you!

PR #585, opened 2026-05-17
Fix idling? blocking on lazy iframes that Chrome never loads

### LINKED ISSUE #583: `idling?` blocks for full `browser.timeout` when lazy iframes never fire `Page.frameStoppedLoading`
**Describe the bug**

When a page contains `<iframe loading="lazy">` elements that Chrome never loads (because they're outside the viewport), Ferrum blocks for the full `browser.timeout` on every page visit.

`idling?` in `ferrum/page/frames.rb` requires every frame registered via `Page.frameAttached` to reach `:stopped_loading`:

```ruby
def idling?
  @frames.values.all? { |f| f.state == :stopped_loading }
end
```

Chrome fires `Page.frameAttached` for every `<iframe>` it discovers in the DOM, including lazy ones it has no intention of loading. It only fires `Page.frameStoppedLoading` when a frame actually completes or fails a load. For a lazy iframe outside the viewport, Chrome parks it indefinitely. No network request, no load event, no `frameStoppedLoading`. Ferrum sits in `go_to`'s wait loop until `browser.timeout` expires.

**To Reproduce**

Two known triggers (also reported in rubycdp/cuprite#303):
1. A `loading="lazy"` iframe inside a closed `<details>` element, present on initial page load.
2. A `loading="lazy"` iframe dynamically inserted outside the viewport via a JS click handler.

`FERRUM_DEBUG=true` makes the issue visible. It streams all CDP messages with elapsed timestamps. In scenario 1, the output shows `Page.frameAttached` firing for two YouTube iframes at ~0.68s, `Page.frameStoppedLoading` firing only for the main frame at ~0.73s, then silence until the 5s timeout.

**Expected behavior**

`go_to` should complete promptly. Lazy iframes that Chrome has chosen not to load should not block the idle check.

**Desktop**

- OS: macOS
- Browser: HeadlessChrome 148.0.0.0
- Version: ferrum 0.17.2, cuprite 0.17

**Additional context**

Both Chrome flag approaches are ineffective on Chrome 148:
- `--disable-features=LazyFrameLoading`
- `--disable-blink-features=LazyFrameLoading`

A workaround for scenario 1 is forcing `loading="eager"` on the iframes in test and using Cuprite's `url_blacklist` to abort the resulting requests. Chrome then fires `frameStoppedLoading` almost immediately on the aborted request. This isn't available for scenario 2 where the iframe is created by JavaScript at runtime.

The most targeted fix would be in the frame subscription logic: a frame that reaches the idle check without ever having fired `Page.frameStartedLoading` (Chrome attached it but never began loading it) could be considered idle rather than stuck. This would handle both scenarios without affecting normal frame lifecycle tracking.

agrberg:
Fixes #583.

## TL;DR

Pages with `loading="lazy"` iframes outside the viewport hang `page.go_to` for the full browser timeout (default 5s) because Chrome attaches such iframes but never starts loading them, leaving `Frame#state` at `nil`. The previous `idling?` required every frame to be `:stopped_loading`, so any nil-state lazy frame blocked idle detection. The fix relaxes the predicate to also accept `nil`.

## Problem

Chrome fires `Page.frameAttached` for every iframe in the DOM, but only fires `Page.frameStoppedLoading` for frames it actually loads. For a `loading="lazy"` iframe parked outside the viewport (for example inside a closed `<details>`), Chrome never starts loading it: no `frameStartedLoading`, no network request, no `frameStoppedLoading`. The frame's `state` stays at `nil` indefinitely.

The previous `idling?` required every frame to be `:stopped_loading`:

```ruby
def idling?
  @frames.values.all? { |f| f.state == :stopped_loading }
end
```

so any lazy iframe blocked idle detection for the full browser timeout. The reporter confirmed via `FERRUM_DEBUG=true` that `Frame#state` was `nil` for the stuck frame.

## Fix

Treat freshly-attached frames (state `nil`) as idle. A frame that has never started loading is not something to wait for.

```ruby
def idling?
  @frames.values.all? { |f| f.state == :stopped_loading || f.state.nil? }
end
```

Semantics: a frame is idle if it has finished loading or has never started loading. This is consistent with `Frame#state`'s documented `nil` initial value and with `subscribe_frame_attached` deliberately leaving state unset on attach.

I considered the alternative of setting `frame.state = :stopped_loading` inside `subscribe_frame_attached`, but that would lie about state (a freshly-attached frame hasn't *stopped* anything) and would break any code that distinguishes "attached but never loaded" from "loaded". The predicate change is documented inline so future cleanup doesn't quietly re-close the issue.

### Behavior change for in-viewport lazy iframes

Ferrum can't tell at idle-check time whether a nil-state frame is one Chrome will never load (the #583 bug) or one Chrome is about to load (an in-viewport lazy iframe whose `frameStartedLoading` is deferred). Treating both as idle is the only correct choice for the never-load case; for the about-to-load case, the consequence is that `page.go_to` may now return before an in-viewport lazy iframe finishes loading. Chrome still loads the iframe; callers that need its content can wait on it explicitly via `frame.body` (which blocks until the frame is loaded) or `page.network.wait_for_idle`. The in-viewport regression test below documents and pins this behavior.

## Test

Five regression tests in `spec/frame_spec.rb`. Each of the four bug-path tests bounds the failure with `with_timeout(1)` so a failing test fails fast, and asserts the operation completes well under that bound.

- **Scenario 1 — initial load**: `loading="lazy"` iframe inside a closed `<details>` on initial page load. Bug surfaces as `go_to` returning after `browser.timeout` (silently rescued, no error raised).
- **Scenario 2 — click handler**: a click handler attaches a lazy iframe and, by design, also triggers a same-document navigation (`location.hash`). The same-document nav is intentional: it causes `idling?` to be re-evaluated while the nil-state frame is in `@frames`. A full reload or cross-document nav here would fire `Runtime.executionContextsCleared` and overwrite the frame state, masking the bug. Bug surfaces as the click hanging until ferrum's internal `mouse_event` raises `Ferrum::TimeoutError`.
- **Reload**: `page.reload` on a page with a lazy iframe. Bug surfaces as `Ferrum::TimeoutError` from `page.reload`.
- **Multiple lazy iframes**: a page with two lazy iframes side by side, mirroring the reporter's real-world case (two YouTube embeds, structurally identical to two `/slow` iframes for the purposes of this bug — the iframe target doesn't matter, only that Chrome defers the load).
- **In-viewport lazy iframe**: pins the behavior change called out above. Asserts that Chrome still loads the iframe and its content is reachable via `frame.body` even though `page.go_to` no longer waits for it.

Each of the four bug-path tests also asserts that a lazy frame in `nil` state is present at the bug's precondition point. This is load-bearing because `Runtime.executionContextsCleared` (in `subscribe_execution_contexts_cleared`) unconditionally overwrites every frame's state to `:stopped_loading`, which on some Chrome versions can fire during navigation and make the old `idling?` return true for the wrong reason. The nil-state assertion catches that false-pass. The reload test captures the precondition *before* `page.reload` runs, because reload itself triggers `executionContextsCleared` and clobbers the state before a post-reload assertion would run; the others capture it after the navigation under test.

Verified each bug-path test fails when the `f.state.nil?` branch is removed from `idling?`.

## Files

- `lib/ferrum/page/frames.rb` — one-line fix plus inline documentation of the nil branch
- `spec/frame_spec.rb` — five regression tests
- `spec/support/views/lazy_iframe.erb` — scenario 1 fixture
- `spec/support/views/lazy_iframe_via_click.erb` — scenario 2 fixture
- `spec/support/views/lazy_iframes_two.erb` — multi-frame fixture
- `spec/support/views/lazy_iframe_in_viewport.erb` — in-viewport behavior-change fixture
- `CHANGELOG.md` — entry under `[Unreleased] → Fixed`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 ### Fixed
 - Full-page screenshots no longer resize the window, preventing focus steal on macOS [#580]
+- `idling?` no longer blocks on `loading="lazy"` iframes that Chrome never starts loading [#585]
 
 ### Removed
 
--- a/lib/ferrum/page/frames.rb
+++ b/lib/ferrum/page/frames.rb
@@ -190,8 +190,25 @@ module Ferrum
         end
       end
 
+      # A frame is idle if it has finished loading (:stopped_loading) or has
+      # never started loading (state == nil).
+      #
+      # The nil case handles loading="lazy" iframes that Chrome attaches but
+      # parks outside the viewport. Chrome creates a Frame in nil state (via
+      # Page.frameAttached or Runtime.executionContextCreated, both of which
+      # call Frame.new and leave state unset), but no frameStartedLoading or
+      # frameStoppedLoading ever follows. Without this branch, idling? blocks
+      # page.go_to for the full browser timeout on any page containing such
+      # an iframe (see #583).
+      #
+      # nil matches the documented initial value on Frame#state and is
+      # structurally unreachable via Frame#state= (which raises ArgumentError
+      # on anything outside STATE_VALUES). Don't "fix" this by initializing
+      # state to :stopped_loading on frameAttached. That would lie about state
+      # ("stopped" implies the frame had loaded) and would silently re-close
+      # #583 with no test failure signal.
       def idling?
-        @frames.values.all? { |f| f.state == :stopped_loading }
+        @frames.values.all? { |f| f.state == :stopped_loading || f.state.nil? }
       end
     end
   end
--- a/spec/frame_spec.rb
+++ b/spec/frame_spec.rb
@@ -125,6 +125,121 @@ describe Ferrum::Frame do
     end
   end
 
+  context "with loading=lazy iframe Chrome never starts loading" do
+    # Regression test for [#583]. Chrome fires Page.frameAttached for every
+    # iframe in the DOM but only fires frameStoppedLoading for frames it
+    # actually loads. A `loading="lazy"` iframe outside the viewport never
+    # starts loading, so its Frame#state stays nil. Without the fix, the
+    # idle predicate refuses to consider such frames idle and page.go_to
+    # blocks for the full browser timeout (silently rescued — no error
+    # is raised, the call just takes ~timeout seconds to return).
+    #
+    # with_timeout(1) bounds the bug path's wall-clock to 1s so a failing
+    # test fails fast. The 0.5s assertion threshold leaves ~5-10x headroom
+    # for the fix path (a static page load is well under 100ms).
+    it "does not block page.go_to (lazy iframe in closed <details>)" do
+      with_timeout(1) do
+        started_at = Ferrum::Utils::ElapsedTime.monotonic_time
+        page.go_to("/lazy_iframe")
+        expect(Ferrum::Utils::ElapsedTime.elapsed_time(started_at)).to be < 0.5
+      end
+
+      # Load-bearing on two axes:
+      # 1. Confirms we hit the bug's precondition (a frame in nil state)
+      #    rather than passing because no lazy frame was attached at all.
+      # 2. Guards against false-pass via Runtime.executionContextsCleared.
+      #    That handler (subscribe_execution_contexts_cleared in
+      #    lib/ferrum/page/frames.rb) unconditionally sets every frame's
+      #    state to :stopped_loading. If it fires during navigation on
+      #    some Chrome versions, the old idling? would return true and
+      #    the timing assertion would pass on `main`. The nil-state
+      #    postcondition catches that case.
+      lazy = page.frames.reject(&:main?).first
+      expect(lazy.state).to be_nil
+    end
+
+    # Scenario 2 from #583. The lazy iframe is inserted by a click handler
+    # AFTER the page has loaded. The handler also triggers a same-document
+    # navigation (location.hash) so idling? is re-evaluated while the
+    # nil-state frame is in @frames. A full reload or cross-document
+    # navigation would fire Runtime.executionContextsCleared and mask the
+    # bug. Under the bug, ferrum's internal mouse_event wait on idling?
+    # blocks for the full browser.timeout and raises Ferrum::TimeoutError.
+    it "does not block click whose handler attaches a lazy iframe" do
+      page.go_to("/lazy_iframe_via_click")
+
+      with_timeout(1) do
+        started_at = Ferrum::Utils::ElapsedTime.monotonic_time
+        page.at_css("#add_iframe").click
+        expect(Ferrum::Utils::ElapsedTime.elapsed_time(started_at)).to be < 0.5
+      end
+
+      # See load-bearing rationale on the scenario 1 test above.
+      lazy = page.frames.reject(&:main?).first
+      expect(lazy.state).to be_nil
+    end
+
+    # Same bug, exercised via reload rather than initial go_to. Under the
+    # bug, page.reload raises Ferrum::TimeoutError after browser.timeout;
+    # with the fix it returns quickly. The post-reload state postcondition
+    # used by the other tests doesn't work here because reload fires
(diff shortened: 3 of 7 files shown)


agrberg:
Pre-review pass on the regression test (force-pushed):

- Replaced the wall-clock 2s threshold with a deterministic predicate test that asserts on `idling?` state directly. No timing assertions — fails on `main`, passes with the fix.
- Added a second scenario from the issue: lazy iframe inserted by JS outside the viewport (the original report's scenario 2). Same code path, but explicit coverage.
- Added the `<link rel="icon" href="data:,">` line to the fixture to match the neighbor iframe fixtures.
- Linked the changelog entry to the PR rather than the issue, matching the surrounding convention.
- Added a comment block above the spec context explaining the Chrome-side root cause so the next reader doesn't have to git-blame.

Diff is still small: one production line + two specs + one fixture + one changelog line.

agrberg:
Tiny review-pass-2 update (force-pushed):

The JS-injected lazy iframe test had a theoretical race — `page.execute`'s CDP ack doesn't order with the `Page.frameAttached` event, so reading `page.frames` immediately after could miss the lazy frame on a slow CI. Wrapped the lookup in `wait_for { ... }.not_to be_nil` (the existing `rspec-wait` idiom used at `spec/page_spec.rb:228`), which polls until the frame appears or hits the 1s wait_timeout. The first scenario (closed `<details>` via `go_to`) was already deterministic — `go_to` waits for the initial frame tree — so it's unchanged.

No production code changes.

agrberg:
Review-pass-3 update (force-pushed): both specs now use only public API — no more `page.send(:idling?)`.

**Approach.** The bug's only externally-observable effect is the wall-clock duration of `page.go_to`: with the bug, `go_to` blocks for the full `browser.timeout` before silently rescuing the resulting `TimeoutError`; with the fix, it returns when the main frame's `frameStoppedLoading` fires. The earlier reviewer-suggested `with_timeout + not_to raise_error` shape doesn't work because `go_to` swallows the timeout. So the test wraps `go_to` in `with_timeout(1)` to bound the bug path to 1s, then uses `Ferrum::Utils::ElapsedTime` (matching the project pattern at `spec/downloads_spec.rb:31`, `spec/network_spec.rb:506`) to assert the call returns well under 0.5s. The fix path is typically <100ms — 5-10x headroom against the threshold. The `lazy.state.nil?` assertion confirms the spec actually hit the bug's precondition rather than passing for an unrelated reason.

**JS-injection scenario.** Reworked: instead of injecting via `page.execute` after `go_to` (which can't reproduce the blocking because by then the main frame is already idle and a subsequent navigation detaches the lazy frame before its `frameStoppedLoading` would matter), there's a new fixture `spec/support/views/lazy_iframe_via_script.erb` that injects the lazy iframe in an inline `<script>` during parse. That gives the same `frameAttached`-before-`frameStoppedLoading` ordering as the closed-`<details>` scenario, just via a different DOM construction path — and the test now genuinely reproduces the bug (verified on `main`: both specs fail at ~1.3s; with the fix they pass at <500ms).

No production code changes.

agrberg:
Similar to #586 I used Claude Code for this and personally reviewed all the code and comments to ensure it takes the least amount of human time to review.

route:
This makes sense to me. I'm going to deep dive today after work hours.

route:
Unfortunately it's not enough to treat nil as iframe being lazy, it's going to generate too many false positives in tests for frames that are barely created and not loaded yet. Also AI is being very verbose, the PR is created that's ok but that makes maintainers to read all these comments, in the code and on github, wasting too much time. Please if you are going to send more PRs try to polish it a bit after AI. My attempt to address this problem is https://github.com/rubycdp/ferrum/pull/597

agrberg:
Thanks for the review, I'll take a look at your implementation as well. Apologies for the chatty AI back-and-forth. I've since changed my approach where I work on a PR with an AI agent on GH and now when it's ready, we open a new clean PR for review.

Issue #584, opened 2026-05-17
`Input.dispatchMouseEvent` fires click before dynamic-import microtasks drain (w/ `element.click()` via `execute_script` workaround)


agrberg:
**Describe the bug**

`Node#click` can fire the browser's `click` event before JavaScript microtasks from dynamic `import()` calls have drained. This means event listeners registered asynchronously like Stimulus action handlers loaded via `eagerLoadControllersFrom` may not yet be connected when the click fires. This causes the handler to silently not run. The test passes on a fast local machine but fails consistently on slower CI.

Replacing the Capybara click with `page.execute_script("arguments[0].click()", element)` reliably fixes this, because `Runtime.evaluate` only executes after all pending microtasks drain.

**To Reproduce**

Given a Stimulus controller loaded via dynamic import:

```javascript
// app/javascript/controllers/example_controller.js
export default class extends Controller {
  copy() { window.__recorded = true }
}
```

```html
<div data-controller="example">
  <button id="btn" data-action="click->example#copy">Click me</button>
</div>
```

```ruby
# Fails on slower CI (Ubuntu), passes locally (macOS):
find("#btn").click
expect(page.evaluate_script("window.__recorded")).to be true

# Always passes:
page.execute_script("arguments[0].click()", find("#btn"))
expect(page.evaluate_script("window.__recorded")).to be true
```

The failure is timing-sensitive. An action wired to `focus->` on a sibling element passes in both environments as `focus` fires synchronously inside Chrome's mousedown processing before the CDP ack returns, so it always fires into a connected listener. A `click->` action does not have this guarantee.

**Expected behavior**

`element.click` should fire into the same listener state as `page.execute_script("arguments[0].click()", element)`. If Ferrum sends `Input.dispatchMouseEvent` and returns, the caller should be able to assume JavaScript event handlers that were scheduled before the call have had a chance to run.

**Desktop (please complete the following information):**

- OS: Fails on Linux (Ubuntu 24.04 CI), passes on macOS 15 (local)
- Browser: Chrome 124+ headless
- Version: Ferrum 0.15, Cuprite 0.17

**Additional context**

[...]

PR #586, opened 2026-05-17
Add opt-in wait_for_pending_js: option to clicks (and FERRUM_WAIT_FOR_PENDING_JS env var)

agrberg:
Refs #584.

## tl;dr

`Mouse#click` returns before pending JS microtasks drain, which causes flaky tests when a click handler depends on a dynamic `import()` resolving (the Stimulus + esbuild pattern). This PR adds an opt-in `wait_for_pending_js:` kwarg on `Mouse#click` and `Node#click`. When set, the click is followed by a no-op `Runtime.evaluate` so the renderer's microtask checkpoint runs before the call returns. `FERRUM_WAIT_FOR_PENDING_JS=true` flips the default project-wide. Default stays `false`, so existing suites see zero behavior change.

Two atomic commits, RBS signatures updated, 13 new specs that spy on the CDP message log.

## Problem

`Mouse#click` issues three `Input.dispatchMouseEvent` CDP commands (mouseMoved, mousePressed, mouseReleased) and returns. The mouseReleased ack returns before the renderer has run microtasks queued behind the click. A common pattern that hits this:

```js
import("./controller").then(({ default: Controller }) => {
  application.register("my-controller", Controller);
});
```

The Stimulus controller's `connect` callback registers the click handler inside the resolved `.then()`. On fast machines the microtask finishes before the next ferrum call; on slower CI the click lands before the handler attaches and the test fails.

The reporter on #584 independently identified the fix shape (no-op `Runtime.evaluate` after the click, which forces the renderer's microtask checkpoint to run before the CDP ack returns).

## Approach

Two atomic commits so the maintainer can land them independently.

### Commit 1: `wait_for_pending_js:` kwarg on `Mouse#click` and `Node#click`

Adds an opt-in kwarg (default `false`). When `true`, follows the click with `Runtime.evaluate(expression: "")`.

```ruby
page.mouse.click(x: x, y: y, wait_for_pending_js: true)
page.at_css("#btn").click(wait_for_pending_js: true)
page.at_css("#btn").click(mode: :right, wait_for_pending_js: true)
page.at_css("#btn").click(mode: :double, wait_for_pending_js: true)
```

`Node#click` now delegates all three modes (`:left`, `:right`, `:double`) to `Mouse#click` so the wait logic lives in one place. `:right` passes `button: :right`, `:double` passes `count: 2`; both pass `wait: 0` to preserve the historical no-network-wait behavior for non-primary clicks (`Mouse#click`'s default `wait: CLICK_WAIT` is preserved for `:left`, which Capybara has historically relied on to block for short navigations). A comment in `Node#click` marks this trade-off.

The flush itself is a private `Mouse#wait_for_pending_js!` method; the public surface is the kwarg on `Mouse#click`. Since `Node#click` delegates, it doesn't need access to the helper.

The kwarg is named after the caller's intent (wait until pending JS settles), not the mechanism (a no-op `Runtime.evaluate` that forces the renderer's microtask checkpoint). If the implementation later switches to `Runtime.awaitPromise` or another barrier, the API still reads correctly.

Opt-in default preserves the historical perf profile for every existing suite. No surprise extra round-trip per click. Users who hit this bug can opt in per call site.

### Commit 2: `FERRUM_WAIT_FOR_PENDING_JS` env var as the default

Mirrors the existing `FERRUM_CLICK_WAIT` / `CLICK_WAIT` pattern at the top of `Mouse` (`mouse.rb:5`). Lets downstream projects (e.g. Cuprite-based suites) flip the default once at boot from a config file instead of plumbing the kwarg through every call site:

```ruby
# config/cuprite.rb (or equivalent; run before the gem loads)
ENV["FERRUM_WAIT_FOR_PENDING_JS"] ||= "true"
require "capybara/cuprite"
```

Exposed as `Ferrum::Mouse::WAIT_FOR_PENDING_JS`. Both `Mouse#click` and `Node#click` default their `wait_for_pending_js:` kwarg to `nil`; `nil` resolves to the constant. Because `Node#click` delegates to `Mouse#click`, the resolution happens in one place. Explicit `true`/`false` at the call site overrides the env-var-flipped default. Specs cover both layers.

The env var is read once at load time, which matches `FERRUM_CLICK_WAIT`. The CHANGELOG entry calls this out so Rails users don't quietly put it in an initializer that runs after `require "ferrum"`.

## Test strategy

The CDP-spy approach is deliberate. An earlier version of this PR asserted on post-click DOM state via `node.text`, but `node.text` goes through `Runtime.callFunctionOn`, which itself drains microtasks. The assertion passed regardless of whether `Mouse#click` issued the extra `Runtime.evaluate`. The test was decorative.

The current specs spy on `Page#command` (the same object `Mouse#click` reaches `@page.command` on) and assert the message log contains or doesn't contain `Runtime.evaluate(expression: "")`. Reverting the conditional flush fails every "must issue" spec.

Default-off specs explicitly `stub_const("Ferrum::Mouse::WAIT_FOR_PENDING_JS", false)` so they pass regardless of the developer's env. The constant-default spec (`be(false)`) gracefully skips when `FERRUM_WAIT_FOR_PENDING_JS=true` is set rather than failing, so developers exercising the opted-in path don't see the ferrum suite break.

## Files

- `lib/ferrum/mouse.rb`: `WAIT_FOR_PENDING_JS` constant, `wait_for_pending_js:` kwarg with `nil` sentinel, private `wait_for_pending_js!` helper.
- `lib/ferrum/node.rb`: all click modes (`:left`, `:right`, `:double`) delegate to `Mouse#click`; `wait_for_pending_js:` threaded through; `:right`/`:double` pass `wait: 0` to preserve historical semantics.
- `sig/ferrum/mouse.rbs`, `sig/ferrum/node.rbs`: RBS signatures updated for the new constant and kwarg (`bool?` on both layers); private helper signature under the `private` section.
- `spec/mouse_spec.rb`: 13 specs covering kwarg-on, kwarg-off (with env guard), all three Node modes, env-var Mouse default, env-var Node default, explicit-false-beats-env at both layers, and a contract spec for the constant's default.
- `CHANGELOG.md`: two entries under `[Unreleased] -> Added`.

## Alternatives considered

**Default-on**. Skipped to avoid changing the perf profile of existing suites without consent. Each click adds one extra CDP round-trip when enabled. Happy to provide measurements if you would prefer default-on.

**Class accessor (`Ferrum::Mouse.wait_for_pending_js_by_default = true`)**. Skipped because ferrum has no existing `.configure` pattern and the env var mirrors `FERRUM_CLICK_WAIT` exactly. Class accessor would work too if you would rather.

**Naming alternatives**. `flush_microtasks:` named the mechanism rather than the intent and would have aged poorly if the implementation later changed to `Runtime.awaitPromise` or another barrier; renamed to `wait_for_pending_js:` for that reason. Open to other names if you have a preference (`wait_for_js:`, `await_js:`, etc.).

## Self-review

Earlier review passes on this PR missed real issues that surfaced on subsequent passes. Posted as a comment for the paper trail and folded into this version. Notable items addressed before this push:

- The two layers used inconsistent semantics for `wait_for_pending_js: nil` (silent-disable on `Mouse`, env-var resolution on `Node`). Both now use the same `nil` sentinel, and `Node#click` delegates the resolution to `Mouse#click` so it only happens in one place.
- `Node#click` originally duplicated the `Runtime.evaluate(expression: "")` string literal that `Mouse#click` also issued. Refactored so `Node#click` delegates all three modes to `Mouse#click`; the helper became private as a free consequence.
- `Node#click`'s env-var integration had no spec.
- Default-off specs were vulnerable to the developer's environment.
- The kwarg was originally named `flush_microtasks:`, which named the mechanism rather than the intent. Renamed to `wait_for_pending_js:` so the API survives an implementation change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### Added
 - `Ferrum::Mouse#click` and `Ferrum::Node#click` accept `wait_for_pending_js:` to drain pending JS microtasks after the click [#586]
+- `Ferrum::Mouse::WAIT_FOR_PENDING_JS` (set by `FERRUM_WAIT_FOR_PENDING_JS`, default `false`) sets the project-wide default for `wait_for_pending_js:`. Read once at load time, so the env var must be set before `require "ferrum"` (e.g. before `require "capybara/cuprite"`) [#586]
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## [Unreleased](https://github.com/rubycdp/ferrum/compare/v0.17.2...main) ##
 
 ### Added
+- `Ferrum::Mouse#click` and `Ferrum::Node#click` accept `wait_for_pending_js:` to drain pending JS microtasks after the click [#586]
 
 ### Changed
 
--- a/lib/ferrum/mouse.rb
+++ b/lib/ferrum/mouse.rb
@@ -66,6 +66,13 @@ module Ferrum
     #
     # @param [Float] wait
     #
+    # @param [Boolean] wait_for_pending_js
+    #   When `true`, follows the click with a no-op `Runtime.evaluate` so the
+    #   renderer's microtask checkpoint runs before this method returns.
+    #   Useful when a click handler depends on state set inside a `.then()`
+    #   chain (for example, a dynamically `import()`-ed Stimulus controller).
+    #   Defaults to `false`, preserving the historical behavior.
+    #
     # @param [Hash{Symbol => Object}] options
     #   Additional keyword arguments.
     #
@@ -79,13 +86,14 @@ module Ferrum
     #
     # @return [self]
     #
-    def click(x:, y:, delay: 0, wait: CLICK_WAIT, **options)
+    def click(x:, y:, delay: 0, wait: CLICK_WAIT, wait_for_pending_js: false, **options)
       move(x: x, y: y)
       down(**options)
       sleep(delay)
       # Potential wait because if some network event is triggered then we have
       # to wait until it's over and frame is loaded or failed to load.
       up(wait: wait, **options)
+      wait_for_pending_js! if wait_for_pending_js
       self
     end
 
@@ -163,6 +171,13 @@ module Ferrum
     end
 
     private
+
+    # Forces the renderer's microtask checkpoint to drain before this call
+    # returns by issuing a no-op `Runtime.evaluate`. Private because the
+    # `wait_for_pending_js:` kwarg on `#click` is the public surface.
+    def wait_for_pending_js!
+      @page.command("Runtime.evaluate", expression: "")
+    end
 
     def mouse_event(type:, button: :left, count: 1, modifiers: nil, wait: 0)
       button = validate_button(button)
--- a/lib/ferrum/mouse.rb
+++ b/lib/ferrum/mouse.rb
@@ -3,6 +3,7 @@
 module Ferrum
   class Mouse
     CLICK_WAIT = ENV.fetch("FERRUM_CLICK_WAIT", 0.1).to_f
+    WAIT_FOR_PENDING_JS = ENV.fetch("FERRUM_WAIT_FOR_PENDING_JS", "false") == "true"
     BUTTON_MASKS = {
       "none" => 0,
       "left" => 1,
@@ -66,12 +67,16 @@ module Ferrum
     #
     # @param [Float] wait
     #
-    # @param [Boolean] wait_for_pending_js
+    # @param [Boolean, nil] wait_for_pending_js
     #   When `true`, follows the click with a no-op `Runtime.evaluate` so the
     #   renderer's microtask checkpoint runs before this method returns.
     #   Useful when a click handler depends on state set inside a `.then()`
     #   chain (for example, a dynamically `import()`-ed Stimulus controller).
-    #   Defaults to `false`, preserving the historical behavior.
+    #
+    #   When `nil` (the default), resolves to {WAIT_FOR_PENDING_JS}, which is
+    #   driven by the `FERRUM_WAIT_FOR_PENDING_JS` env var. Set the env var
+    #   to `"true"` before requiring ferrum to flip the project-wide default.
+    #   It is read once at load time.
     #
     # @param [Hash{Symbol => Object}] options
     #   Additional keyword arguments.
@@ -86,14 +91,15 @@ module Ferrum
     #
     # @return [self]
     #
-    def click(x:, y:, delay: 0, wait: CLICK_WAIT, wait_for_pending_js: false, **options)
+    def click(x:, y:, delay: 0, wait: CLICK_WAIT, wait_for_pending_js: nil, **options)
+      should_wait = wait_for_pending_js.nil? ? WAIT_FOR_PENDING_JS : wait_for_pending_js
       move(x: x, y: y)
       down(**options)
       sleep(delay)
       # Potential wait because if some network event is triggered then we have
       # to wait until it's over and frame is loaded or failed to load.
       up(wait: wait, **options)
-      wait_for_pending_js! if wait_for_pending_js
+      wait_for_pending_js! if should_wait
       self
     end
 
--- a/lib/ferrum/node.rb
+++ b/lib/ferrum/node.rb
@@ -64,6 +64,9 @@ module Ferrum
     # keys: (:alt, (:ctrl | :control), (:meta | :command), :shift)
     # offset: { :x, :y, :position (:top | :center) }
     # wait_for_pending_js: see Mouse#click. Applies to all click modes.
+    #   Default `nil` resolves to Mouse::WAIT_FOR_PENDING_JS so a single
+    #   FERRUM_WAIT_FOR_PENDING_JS setting covers both layers. Explicit
+    #   true/false at the call site wins over the env-var default.
     #
     # All three modes delegate to Mouse#click so `wait_for_pending_js:` lives
     # in one place. The `:right` and `:double` modes pass `wait: 0` to
@@ -71,7 +74,7 @@ module Ferrum
     # non-primary buttons; `:left` uses Mouse#click's default `wait: CLICK_WAIT`
     # because Capybara has historically relied on the primary click to block
     # for short navigations.
-    def click(mode: :left, keys: [], offset: {}, delay: 0, wait_for_pending_js: false)
+    def click(mode: :left, keys: [], offset: {}, delay: 0, wait_for_pending_js: nil)
       x, y = find_position(**offset)
       modifiers = page.keyboard.modifiers(keys)
 
--- a/lib/ferrum/node.rb
+++ b/lib/ferrum/node.rb
@@ -63,23 +63,28 @@ module Ferrum
     # mode: (:left | :right | :double)
     # keys: (:alt, (:ctrl | :control), (:meta | :command), :shift)
     # offset: { :x, :y, :position (:top | :center) }
-    def click(mode: :left, keys: [], offset: {}, delay: 0)
+    # wait_for_pending_js: see Mouse#click. Applies to all click modes.
+    #
+    # All three modes delegate to Mouse#click so `wait_for_pending_js:` lives
+    # in one place. The `:right` and `:double` modes pass `wait: 0` to
+    # preserve the historical no-network-wait behavior of `Node#click` for
+    # non-primary buttons; `:left` uses Mouse#click's default `wait: CLICK_WAIT`
+    # because Capybara has historically relied on the primary click to block
+    # for short navigations.
+    def click(mode: :left, keys: [], offset: {}, delay: 0, wait_for_pending_js: false)
       x, y = find_position(**offset)
       modifiers = page.keyboard.modifiers(keys)
 
       case mode
       when :right
-        page.mouse.move(x: x, y: y)
-        page.mouse.down(button: :right, modifiers: modifiers)
-        sleep(delay)
-        page.mouse.up(button: :right, modifiers: modifiers)
+        page.mouse.click(x: x, y: y, button: :right, modifiers: modifiers, delay: delay,
+                         wait: 0, wait_for_pending_js: wait_for_pending_js)
       when :double
-        page.mouse.move(x: x, y: y)
-        page.mouse.down(modifiers: modifiers, count: 2)
-        sleep(delay)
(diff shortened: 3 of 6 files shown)


agrberg:
Pre-review pass on the test infrastructure and the right/double-click coverage (force-pushed).

**Blocker addressed**: the previous specs asserted on post-click DOM state via `node.text`, which itself goes through `Runtime.*` and drains microtasks, so they passed regardless of the new code path. Replaced with focused specs that spy on `Page#command` and assert that `Runtime.evaluate` is or isn't issued after the click depending on the option. Verified all five "must issue" specs fail on a build with the conditional flush removed.

**Decoupled Node from `Mouse::WAIT_FOR_PENDING_JS`**: `Node#click` now takes `wait_for_pending_js: nil` and resolves to the Mouse constant inline, rather than referencing it as the signature default. Less surprising to read, and `nil` clearly signals "defer to default."

**Extended to `:right` and `:double` modes** in `Node#click`. Same CDP race applies. If we're shipping the kwarg, it should cover all click paths. Each mode has its own spec.

**Tighter CHANGELOG entries** to match neighbor style (see the analogous `CLICK_WAIT` entry for shape).

**Re: opt-in vs default-on**: kept opt-in. The kwarg overhead is one `Runtime.evaluate` round-trip per click. Happy to provide measurements if you would prefer default-on, or to flip the default and ship default-on as a third commit; I went conservative to avoid changing the perf profile of every existing suite without consent.

Dropped the `click_after_microtask.erb` fixture since the spy specs use the existing `/click_coordinates` fixture and do not need a microtask-flavored page.

(Note: the kwarg was originally proposed as `flush_microtasks:` and was renamed to `wait_for_pending_js:` in a later force-push to name the caller's intent rather than the mechanism. Identifiers above use the current name.)

agrberg:
Review-pass-2 update (force-pushed both commits). All changes are additive. No production logic changes; only fills in gaps that a pre-submission review flagged.

**1. RBS signatures updated** (the most important one). The kwarg was missing from `sig/ferrum/mouse.rbs` and `sig/ferrum/node.rbs`, and the `WAIT_FOR_PENDING_JS` constant was missing from the Mouse signature. Added in both commits where the corresponding Ruby surface lands:
- Commit 1 adds `?wait_for_pending_js: bool` to `Mouse#click` and `Node#click`.
- Commit 2 adds `WAIT_FOR_PENDING_JS: bool` and widens the kwarg type to `bool?` on both layers to match the `nil` sentinel.

Verified with `bundle exec rbs -I sig validate`. No ferrum-related errors.

**2. New spec: explicit kwarg beats env-var default.** Closes a composition gap noted in pre-submission review: previously no spec asserted that `wait_for_pending_js: false` at the call site overrides an env-var-flipped default. `stub_const WAIT_FOR_PENDING_JS=true` + `click(wait_for_pending_js: false)` expects no `Runtime.evaluate`.

**3. PR description refresh.** Removed a dead reference to a fixture file (`click_after_microtask.erb`) that was dropped in the test rewrite. Now reflects the actual diff.

**4. Doc clarification.** Added a one-liner to the `wait_for_pending_js:` doc and to the second commit's message noting that the env var is read once at require time (use `stub_const` for runtime mutation).

131 specs pass locally, RuboCop clean, RBS validates clean.

Open questions still listed in the PR description for you to decide:
- Default-on vs opt-in
- Class accessor vs env var as the project-wide knob

I've defaulted conservatively on both; happy to flip either.

(Note: the kwarg was originally `flush_microtasks:` and was renamed to `wait_for_pending_js:` in a later force-push. Identifiers above use the current name.)

agrberg:
Self-review pass turned up the following at the time it was posted. **All items below are addressed in the current diff.** Kept here for the paper trail; identifier names updated to match the current code (kwarg `wait_for_pending_js:`, constant `WAIT_FOR_PENDING_JS`, env var `FERRUM_WAIT_FOR_PENDING_JS`, private helper `Mouse#wait_for_pending_js!`).

## Code

1. **`Node#click` duplicated the wait mechanism instead of delegating to `Mouse`.** Two paths (`Mouse#click` and `Node#click` for `:right`/`:double`) issued the same `@page.command("Runtime.evaluate", expression: "")` literal independently. If the mechanism ever changed (e.g. to `Runtime.awaitPromise`), the right/double paths would silently keep using `Runtime.evaluate`. **Fixed** by refactoring `Node#click` so all three modes delegate to `Mouse#click`. The wait helper became private as a free consequence: `Node` no longer needs external access to it.

2. **`Mouse#click(wait_for_pending_js: nil)` silently disabled the wait.** With `if wait_for_pending_js`, `nil` was falsy and did nothing. But `Node#click(wait_for_pending_js: nil)` resolved `nil` to `Mouse::WAIT_FOR_PENDING_JS`. Same kwarg, opposite semantics across the two layers. The RBS asymmetry (`bool` on Mouse, `bool?` on Node) was the type-system tell. **Fixed** by making `Mouse#click` use the same `nil` sentinel and widening RBS on both layers to `bool?`. Because `Node#click` now delegates, the resolution happens in `Mouse#click` only.

3. **The `:left` early-return in `Node#click` was load-bearing but undocumented.** **Fixed** by eliminating the asymmetry entirely: all three modes now delegate to `Mouse#click`. `Node#click` has no post-case branch and no special `:left` return. A comment explains why `:right`/`:double` pass `wait: 0` (preserves the historical no-network-wait behavior for non-primary clicks).

## Test coverage

4. **`Node#click`'s env-var integration was untested.** The `stub_const("Ferrum::Mouse::WAIT_FOR_PENDING_JS", true)` specs only exercised `Mouse#click`. Replacing the Node-layer resolution with `false` during a refactor would have still passed all eight existing specs. **Fixed** by adding a Node-layer env-var spec.

5. **The "default `false`" specs depended on an implicit clean env.** They assumed `FERRUM_WAIT_FOR_PENDING_JS` was unset. Cuprite-using developers who follow the very pattern this PR documents would have seen them fail with no hint why. **Fixed** by adding a `stub_const("Ferrum::Mouse::WAIT_FOR_PENDING_JS", false)` precondition so the specs are deterministic regardless of env. The contract spec that asserts the default is `false` gracefully skips when the env var is set rather than failing.

## Maintainability

6. **Env var read at require time is a Rails ordering footgun.** `mouse.rb` evaluates `ENV.fetch("FERRUM_WAIT_FOR_PENDING_JS", ...)` once. The recommended `ENV[...] ||= "true"; require "capybara/cuprite"` pattern only works if the env var is set before ferrum loads. A Rails initializer that runs after auto-load silently no-ops. Matches the existing `CLICK_WAIT` convention, so defensible, but the commit message noted it while user-facing docs did not. **Fixed** by adding a note to the changelog entry.

## Copy

7. **PR body and YARD comment used em-dashes.** **Fixed** by rewriting to plain prose.

8. **CHANGELOG line was awkward.** Original phrasing used "as" twice in one sentence. **Fixed** by rewording.

## Naming

9. **Kwarg was originally named `flush_microtasks:`, which named the mechanism rather than the intent.** If the implementation later switches to `Runtime.awaitPromise` or another barrier, the name would age poorly. **Fixed** by renaming to `wait_for_pending_js:` across all surfaces (kwarg, helper, constant, env var). PR description, commit messages, and related comments updated to match.

## Structure

10. **`wait_for_pending_js!` was public because `Node#click` reached in for `:right`/`:double`.** **Fixed** by refactoring `Node#click` to delegate all modes to `Mouse#click` (with `wait: 0` on `:right`/`:double` to preserve historical semantics). The helper is now private; the public surface is just the kwarg.

agrberg:
Wanted to make explicit something that may not be obvious from the diff: refactoring `Node#click`'s `:right` and `:double` branches to delegate to `Mouse#click` is purely encapsulation. The CDP message stream is byte-identical to the previous inline form.

Verified by spying on `Page#command` during both forms and comparing the captured `Input.dispatchMouseEvent` payloads (run on a detached worktree off this branch, default env):

**`:right`**

```
OLD inline (move + down(button: :right) + up(button: :right)):
  {type: "mouseMoved",    x: 100.0, y: 150.0, buttons: 0}
  {type: "mousePressed",  x: 100,   y: 150,   buttons: 2, clickCount: 1, button: "right", modifiers: 0, wait: 0}
  {type: "mouseReleased", x: 100,   y: 150,   buttons: 0, clickCount: 1, button: "right", modifiers: 0, wait: 0}

NEW Mouse#click(button: :right, wait: 0):  IDENTICAL
```

**`:double`**

```
OLD inline (move + down(count: 2) + up(count: 2)):
  {type: "mouseMoved",    x: 100.0, y: 150.0, buttons: 0}
  {type: "mousePressed",  x: 100,   y: 150,   buttons: 1, clickCount: 2, button: "left", modifiers: 0, wait: 0}
  {type: "mouseReleased", x: 100,   y: 150,   buttons: 0, clickCount: 2, button: "left", modifiers: 0, wait: 0}

NEW Mouse#click(count: 2, wait: 0):  IDENTICAL
```

**`:left`** (no semantic change; same call shape with new kwarg defaulting through):

```
OLD Mouse#click(x:, y:, modifiers:, delay:):
  {type: "mouseMoved",    x: 100.0, y: 150.0, buttons: 0}
  {type: "mousePressed",  x: 100,   y: 150,   buttons: 1, clickCount: 1, button: "left", modifiers: 0, wait: 0}
  {type: "mouseReleased", x: 100,   y: 150,   buttons: 0, clickCount: 1, button: "left", modifiers: 0, wait: 0.1}

NEW Mouse#click(..., wait_for_pending_js: nil):  IDENTICAL
```

The only observable change is when a caller explicitly opts in via `wait_for_pending_js: true` (or sets `FERRUM_WAIT_FOR_PENDING_JS=true`), which appends one `Runtime.evaluate` call after the click sequence. That's the feature the PR is adding.

One pre-existing behavior this PR deliberately does **not** change: `:right`/`:double` skip the network wait on `mouseReleased`. `Mouse#click` normally passes `wait: CLICK_WAIT` (0.1s) to `up()`, but the inline form in `Node#click` historically didn't, so I pass `wait: 0` to `Mouse#click` for those modes to preserve that. See the comment in `node.rb` above the `case`.

agrberg:
I used Claude Code to diagnose, workaround, and help me write up #584 in addition to writing and reviewing this PR. I've personally reviewed all the code and comments and believe it ready for human review at this point.

route:
I don't like it and it doesn't belong to ferrum. Whether or not you expect something from click, it's not going to change in ferrum, click is as simple as click.

agrberg:
I pulled out the [small readability clean up](https://github.com/rubycdp/ferrum/pull/586#issuecomment-4472280101) into #587 . It's not going to win any awards 😅 but I think it'll help the next person reading the method.

PR #587, opened 2026-05-18
`Node#click` reusability of `Mouse#click`

agrberg:
Increase `Node#click` readability by making the differences between modes obvious when to `Mouse#click`'s encapsulation

Further focus attention on only the changes by using Ruby 3.1+ kwarg shorthand.

### DIFF
--- a/lib/ferrum/node.rb
+++ b/lib/ferrum/node.rb
@@ -67,19 +67,15 @@ module Ferrum
       x, y = find_position(**offset)
       modifiers = page.keyboard.modifiers(keys)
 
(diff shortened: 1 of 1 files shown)


agrberg:
FWIW the Ruby 3.1+ kwarg shorthand isn't necessary so I'll easily remove if it you don't want to use it in this repo. I added it as the [gemspec](https://github.com/rubycdp/ferrum/blob/main/ferrum.gemspec#L26) says Ruby 3.1+ is required and omitting it draws the eye to only what is changing.

route:
looking good! yes we should support only 3.1+

PR #595, opened 2026-06-27
feat: add `page.accessibility` CDP domain API

myabc:
## Problem

Reading an element's computed accessibility properties (role, name, description, ARIA state) under Ferrum means reaching past the public API into raw CDP:

```ruby
page.command("Accessibility.getPartialAXTree",
  nodeId: node.node_id, fetchRelatives: false)["nodes"]
```

This leaks the CDP wire shape into callers, is undiscoverable, and leaves no home for the rest of the Accessibility domain.

## Fix

Add a first-class `page.accessibility` domain object, mirroring `Ferrum::Network`:

- `node_for`, `partial_tree`, `snapshot`, `query`, `root`, `enable`/`disable`. Node-scoped commands run on the node's own page, so they resolve correctly across same-process iframes.
- `Ferrum::Accessibility::AXNode` value object wraps a CDP AXNode: `role`, `name`, `description`, `value`, `properties`, `ignored?`, `ignored_reasons`, `node_id`, `backend_dom_node_id`, `child_ids`, `to_h`. Params are deep-frozen; `value` is `untyped` since CDP `AXValue.value` may be a string, number, or boolean.
- `Ferrum::Node#axnode` convenience, mirroring `#computed_style`.

Query commands work without `Accessibility.enable` (that governs live events only); `enable`/`disable` are thin wrappers. `root` uses `getFullAXTree(depth: 1)` and `query` falls back to the document root node id, because CDP `getRootAXNode`/`queryAXTree` require those respectively.

## Testing

RBS signatures plus unit and browser-driven specs, including subtree-scoping proof and same-process iframe coverage for `node_for`, `partial_tree`, and `query`. CHANGELOG updated.

## Downstream validation

🤖 Validated against a real consumer: [`capybara_accessible_selectors#157`](https://github.com/citizensadvice/capybara_accessible_selectors/pull/157) (Cuprite driver support) resolves computed `name`/`role`/`description` from the accessibility tree.

Swapping its hand-rolled `Accessibility.getPartialAXTree` + ignored-node filtering + hash-digging for `node.axnode&.name` / `.role` / `.description` removed the raw-CDP coupling entirely. Full suite green on Cuprite — **1813 examples, 0 failures**.

Before:

```ruby
nodes = node.page.command("Accessibility.getPartialAXTree",
                          nodeId: node.node_id, fetchRelatives: false)["nodes"]
Array(nodes).find { |n| n["ignored"] == false }&.dig("name", "value")
```

After:

```ruby
node.axnode&.name
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - `Ferrum::Frame#loader_id` provides a loader id when frame navigates [#583]
 - `Ferrum::Frame#lifecycle_events` provides a list of frame's events like init, networkIdle, firstPaint, etc. [#583]
 - `Ferrum::Frame#idle?` whether frame was loaded [#583]
+- `page.accessibility` API and `Node#axnode` for reading the CDP accessibility tree
 
 ### Changed
 - `Ferrum::PendingConnectionsError` and `Ferrum::TimeoutError` were swallowed even though happening when traffic iterator results in empty array. [#583]
--- a/lib/ferrum/accessibility.rb
+++ b/lib/ferrum/accessibility.rb
@@ -1,0 +1,102 @@
+# frozen_string_literal: true
+
+require "ferrum/accessibility/ax_node"
+
+module Ferrum
+  #
+  # Wraps the CDP [Accessibility](https://chromedevtools.github.io/devtools-protocol/tot/Accessibility/)
+  # domain. The query commands work without `enable`; `enable`/`disable` are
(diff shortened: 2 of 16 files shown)


route:
Interesting, requires a thorough review

myabc:
@route let me know if you have any questions! I'm pretty open to suggestions - particularly with regards to naming.

route:
Perfect! Thanks again!

PR #596, opened 2026-06-29
fix: Text nodes are not what they seem

route:
Send all nodes from backend to frontend including hidden whitespace nodes which resolve to node id 0 without this setting on

https://github.com/rubycdp/ferrum/discussions/594

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,13 @@
 
 ### Added
 
+
 ### Changed
 
 ### Fixed
 - Full-page screenshots no longer resize the window, preventing focus steal on macOS [#580]
+- DOM.enable is now has `includeWhitespace: "all"` to keep track of new line nodes which previously were resolved to 0, and errored with NodeNotFoundError [#596]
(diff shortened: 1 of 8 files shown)


Nakilon:
For context.

Ferrum may throw `Could not find node with given id (Ferrum::NodeNotFoundError)` on trying to access "html-ignorable whitespace" text nodes like this: `.xpath(".//text()")`.  
`<div><span> </span></div>` -- this is ok  
`<div> <span></span></div>` -- this is exception  
You can't the see cause in Elements tab because this whitespace isn't essentially shown in the DOM tree.

Also when you open local html files, Chrome adds `\n` before `</body>` to create the issue no matter what, like this:
```ruby
"<html lang=\"ru\"><head></head><body><div>text1<span>text2</span></div>\n</body></html>"
```

Workarounds were:  
1. export outerHTML of the node and parse with Nokogiri  
2. force removal of those whitespaces:
    ```ruby
    listitem.evaluate <<~JS
      (function(){
        const nodes = [];
        const walker = document.createTreeWalker(this, NodeFilter.SHOW_TEXT);
        let node
        while (node = walker.nextNode())
          if (!node.nodeValue.trim())
            nodes.push(node);
        nodes.forEach(n => n.remove());
      }).call(this)
    JS

```

PR #598, opened 2026-07-06
Make webrick dependency optional

odlp:
## What

This PR proposes making the `webrick` dependency optional, by removing it from the gemspec. It's only used by `Ferrum::Proxy`.

This is a breaking change, so user's would have to add `webrick` to their Gemfile if it's not already present.

## Why

Fewer dependencies in our application means less time updating them & less surface area for supply chain attacks / vulnerabilities.

This PR was motivated by the recent CVE found in `webrick`:

https://nvd.nist.gov/vuln/detail/cve-2026-38969

## Notes

Unfortunately Rubygems / Bundler doesn't have an API for specifying optional dependencies:
https://github.com/ruby/rubygems/issues/1811

But the `rescue LoadError` / informative error message pattern seems fairly widespread as a method of handling optional dependencies in gems.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ### Changed
 - `Ferrum::PendingConnectionsError` and `Ferrum::TimeoutError` were swallowed even though happening when traffic iterator results in empty array. [#583]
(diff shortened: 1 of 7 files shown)


route:
Looking good, thanks!

PR #600, opened 2026-07-09
fix: trigger select-all via CDP commands field

myabc:
## Problem

`Ctrl`/`Cmd`+`A` does not select text when driven through `Ferrum::Node#type` / `Ferrum::Keyboard`. Selecting-then-replacing content in an `<input>` or a `contenteditable` is impossible, so callers resort to workarounds. Ferrum's own specs sidestep it — the `delete_all` helper in `spec/node_spec.rb` selects via `Shift`+modifier+`Right` caret movement rather than `Ctrl`+`A`, which only works for single-word content.

## Root cause

Synthetic key events dispatched via `Input.dispatchKeyEvent` do not trigger the browser's editing accelerators. This holds regardless of `text` presence or `keyDown`/`rawKeyDown` type — verified by probing a live browser: no key-event variant of `Ctrl`/`Cmd`+`A` produces a selection.

The reliable mechanism is the CDP [`commands`](https://chromedevtools.github.io/devtools-protocol/tot/Input/#method-dispatchKeyEvent) field on the key event: passing `commands: ["selectAll"]` makes Chrome run the editing command. This is how Playwright and Puppeteer implement editing shortcuts.

## Fix

When a chord matches the platform select-all accelerator (`Cmd` on macOS, `Ctrl` elsewhere), name the editing command via `commands`. Existing `text`/`modifiers` handling is untouched — only an extra field is added for the matching chord, so normal typing and all existing modifier behaviour are unaffected.

The `editing_command` helper is scoped to select-all for now (the common need) and excludes `Shift`, but is written to extend to further commands.

## Testing

New `#type` specs in `spec/node_spec.rb` cover a `contenteditable` (`#filled_div`) and an `<input>` (`#filled_input`): `type([mod, "a"], :backspace)` clears the field. The full `#type` context stays green (44 examples).

## Real-world beneficiaries

- [capybara_accessible_selectors](https://github.com/citizensadvice/capybara_accessible_selectors) — clearing a rich-text `contenteditable` currently needs a driver-specific backspace loop.
- [OpenProject's Capybara helpers](https://github.com/opf/openproject/blob/dev/spec/support/shared/cuprite_helpers.rb#L66) — `clear_input_field_contents` does a manual right+backspace dance on inputs (commented "Ferrum is yet support `fill_options`…"); collapses to `node.type([:control, "a"], :backspace)` once the accelerator works.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 - `Ferrum::Frame#lifecycle_events` provides a list of frame's events like init, networkIdle, firstPaint, etc. [#583]
 - `Ferrum::Frame#idle?` whether frame was loaded [#583]
 
+### Fixed
+- `Ferrum::Node#type` / `Ferrum::Keyboard` now trigger the select-all editing shortcut (`Ctrl`/`Cmd`+`A`) by naming the CDP `commands` field, so selecting and replacing text in inputs and contenteditables works.
+
 ### Changed
 - `Ferrum::PendingConnectionsError` and `Ferrum::TimeoutError` were swallowed even though happening when traffic iterator results in empty array. [#583]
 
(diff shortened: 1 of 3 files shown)


route:
Thank you!

PR #603, opened 2026-07-13
Log websocket responses before delivering messages

kanutocd:
## Summary

- Log websocket responses before delivering them to waiting commands
- Prevent callers from reading logger output before the command response has been written

## Verification
- `bundle exec rspec spec/unit/browser_spec.rb`
- `bundle exec rspec spec/browser_spec.rb:44`
- `bundle exec rubocop lib/ferrum/client/web_socket.rb spec/unit/browser_spec.rb`
- `bundle exec rake test`

### DIFF
--- a/lib/ferrum/client/web_socket.rb
+++ b/lib/ferrum/client/web_socket.rb
@@ -51,12 +51,6 @@ module Ferrum
 
       def on_message(event)
         data = safely_parse_json(event.data)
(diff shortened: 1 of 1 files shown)


route:
That's not a race per se, but ok! You are not moving a comment along

kanutocd:
@route Thanks, that makes sense. I updated the PR 's title to avoid framing this as a race, and moved the comment together with the `@messages.push(data) if data` line so it still documents the code it refers to.

PR #608, opened 2026-08-05
Fix RGBA spec filename typo

i7a:
## Summary

  Follow-up to #286, which renamed `lib/ferrum/rbga.rb` to `lib/ferrum/rgba.rb`.

  The corresponding spec filename was left unchanged, so this renames `spec/rbga_spec.rb` to `spec/rgba_spec.rb`. No test contents were
  changed.

  ## Verification

  - `bundle exec rspec spec/rgba_spec.rb` — 7 examples, 0 failures
  - `bundle exec rubocop --cache false` — 97 files inspected, no offenses


route:
Thanks!

PR #611, opened 2026-08-09
fix: use backend node id for file inputs

### LINKED ISSUE #568: DOM.setFileInputFiles crashes renderer when using nodeId
## Problem

Calling `Node#select_file` can crash the browser tab with:

```
ERROR:content/browser/bad_message.cc:29  Terminating renderer for bad IPC message, reason 2
```

Reason 2 corresponds to `RFH_CAN_ACCESS_FILES_OF_PAGE_STATE` — Chrome's security check that verifies the renderer process has been granted read access to files referenced in page state.

## Root cause

Ferrum's `Node#select_file` uses `nodeId` to identify the element:

```ruby
# lib/ferrum/node.rb:113
page.command("DOM.setFileInputFiles", slowmoable: true, nodeId: node_id, files: Array(value))
```

`DOM.setFileInputFiles` with `nodeId` can fail to properly grant the renderer process read access to the specified files. When Chrome later serializes page state containing the file reference, the access check fails and the renderer is killed.

This is particularly reproducible after same-document navigations (e.g. Turbo Drive in Rails apps), where the `nodeId` may reference a stale DOM context.

## Fix

Using `backendNodeId` instead of `nodeId` resolves the crash. The `backendNodeId` is already available via `description["backendNodeId"]`:

```ruby
def select_file(value)
  files = Array(value).map { |f| File.expand_path(f) }
  page.command("DOM.setFileInputFiles", slowmoable: true,
                                        backendNodeId: description["backendNodeId"],
                                        files: files)
end
```

For reference, Puppeteer uses `objectId` (not `nodeId`) for this CDP command, which is why Puppeteer-based tools don't hit this issue.

## Environment

- Ferrum: 0.17.1
- Cuprite: 0.17
- Chromium: 145.0.7632.45
- macOS (Darwin 25.3.0)
- Ruby 4.0.1

oiahoon:
## Summary

- identify file inputs with their stable `backendNodeId` when calling `DOM.setFileInputFiles`
- cover both the exact CDP command and a real Chrome file selection

Fixes #568

## Verification

- `bundle exec rspec spec/node_spec.rb --example '#select_file'` (2 examples)
- `bundle exec rake test` (556 examples; the two local Chrome/path failures reproduce on unchanged `main`)
- `bundle exec rubocop --parallel` (99 files)
- `bundle exec rake build`
- `git diff --check`

### DIFF
--- a/lib/ferrum/node.rb
+++ b/lib/ferrum/node.rb
@@ -121,7 +121,12 @@ module Ferrum
     end
 
     def select_file(value)
-      page.command("DOM.setFileInputFiles", slowmoable: true, nodeId: node_id, files: Array(value))
+      page.command(
+        "DOM.setFileInputFiles",
+        slowmoable: true,
+        backendNodeId: description["backendNodeId"],
+        files: Array(value)
+      )
     end
 
     def at_xpath(selector)
--- a/spec/node_spec.rb
+++ b/spec/node_spec.rb
@@ -861,6 +861,37 @@ describe Ferrum::Node do
     end
   end
 
+  describe "#select_file" do
+    it "selects a file" do
+      browser.go_to("/form")
+      input = browser.at_css("#form_image")
+
+      input.select_file(File.expand_path(__FILE__))
+
+      expect(input.value).to eq("C:\\fakepath\\node_spec.rb")
+    end
+
+    it "uses the backend node id for the file input" do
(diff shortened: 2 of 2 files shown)


route:
Thanks! I'll merge and remove unit test, it's a spec for real behavior not unit tests.

PR #616, opened 2026-08-20
feat: improve process termination and directory cleanup logic

route:
- Add `send_signal` method for robust PID and process group signaling.
- Implement `remove_directory` with retries and backoff to handle transient errors.

### DIFF
--- a/lib/ferrum/browser/process.rb
+++ b/lib/ferrum/browser/process.rb
@@ -58,12 +58,21 @@ module Ferrum
           else
             send_signal(pid, "TERM")
             start = Utils::ElapsedTime.monotonic_time
-            while ::Process.wait(pid, ::Process::WNOHANG).nil?
+            leader_exited = false
+            loop do
+              # The leader (the pid we actually spawned and can #wait on) may
+              # exit well before the rest of its process group does -- e.g. a
+              # renderer/zygote that ignores TERM. Reaping the leader must not
+              # by itself end the loop, or that child is orphaned forever
+              # since KILL is only ever sent from the timeout branch below.
+              leader_exited ||= !::Process.wait(pid, ::Process::WNOHANG).nil?
+              break if leader_exited && !process_group_alive?(pid)
+
               sleep(WAIT_KILLED)
               next unless Utils::ElapsedTime.timeout?(start, KILL_TIMEOUT)
 
               send_signal(pid, "KILL")
-              ::Process.wait(pid)
+              ::Process.wait(pid) unless leader_exited
               break
             end
           end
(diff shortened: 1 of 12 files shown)


lib/ferrum/browser/process.rb
```diff
@@ -54,13 +56,13 @@ def self.process_killer(pid)
             # Process.kill is unreliable on Windows
             ::Process.kill("KILL", pid) unless system("taskkill /f /t /pid #{pid} >NUL 2>NUL")
           else
-            ::Process.kill("USR1", pid)
+            send_signal(pid, "TERM")
             start = Utils::ElapsedTime.monotonic_time
             while ::Process.wait(pid, ::Process::WNOHANG).nil?
```
ZilvinasKucinskas: This returns as soon as the process-group leader exits. If Chrome exits on `TERM` while a renderer or zygote stays alive, `wait` reaps the leader and group `KILL` is never sent. I reproduced this on macOS and Debian with a child in the same group that ignores `TERM` (`child_alive=true` on this commit).

Please track leader reaping separately and keep polling the group until it exits or the timeout sends `KILL`:

```ruby
leader_exited = false
loop do
  leader_exited ||= !::Process.wait(pid, ::Process::WNOHANG).nil?
  break if leader_exited && !process_group_alive?(pid)

  sleep(WAIT_KILLED)
  next unless Utils::ElapsedTime.timeout?(start, KILL_TIMEOUT)

  send_signal(pid, "KILL")
  ::Process.wait(pid) unless leader_exited
  break
end
```

`process_group_alive?` can use `Process.kill(0, -pid)`, treating `ESRCH` as false and `EPERM` as true. Please add one real subprocess regression; the current mocks always return `nil` and miss this case.

lib/ferrum/browser/process.rb
```diff
@@ -79,13 +100,38 @@ def self.process_killer(pid)
...
+      #   Directory to remove.
+      # @param [Integer] retries
+      #   Maximum number of removal attempts.
+      # @param [Float] delay
+      #   Base delay, in seconds, before the first retry; doubles on each
+      #   subsequent attempt.
+      #
+      # @return [void]
+      #
+      def self.remove_directory(path, retries: REMOVE_DIR_RETRIES, delay: REMOVE_DIR_RETRY_DELAY)
```
ZilvinasKucinskas: Optional: with the 2s browser timeout plus 1.5s directory backoff, `quit` can block for about 3.5s, or 5.5s with Xvfb. Could we expose an opt-in `Browser#quit(wait: false)` / `Process#stop(wait: false)`?

Please keep `wait: true` as the default and keep `restart` synchronous, so Ferrum and Cuprite behavior stays unchanged. The async path should close CDP synchronously, run browser group -> Xvfb -> user-data-directory cleanup through `Utils::Thread.spawn(abort_on_exception: false)`, and return the thread or handle so callers can join it before process exit or fixed-port reuse. We currently carry this as an application patch and could remove it.

ZilvinasKucinskas (changes requested):
I tested this head against real process groups on macOS and Debian. The blocking issue below reproduces on the current code and passes with the suggested loop. The candidate also passes Ferrum's full suite and lint; the async API note is optional.

lib/ferrum/browser/process.rb
route: thats an oversight likely from my side, looking at your patch gist. ported it over too.

PR #619, opened 2026-08-28
fix: replace no-op `--no-crashpad` with `--disable-crashpad-for-testing`

route:
`--no-crashpad`, added to the default flags in 0.18.0 to stop `chrome_crashpad_handler` from spawning, is not a Chromium switch. Chrome silently ignores switches it doesn't recognise, so two handler processes kept starting per browser.

Measured on Chrome 127, 128, 152 and 154 by launching each build with and without the flag and counting handlers: identical either way. The real switch is `disable-crashpad-for-testing`, defined in `chrome/common/chrome_switches.cc`, which takes two handlers to zero.

This only became reachable in Chrome 128. Before that, bare `--headless` selected old headless, a separate lightweight shell that never started a crash handler; from 128 it selects new headless, which is full Chrome. The switch was added in that same release, so it covers exactly the range where the handlers appear.

macOS is unaffected either way: one handler starts regardless of the flag, and exits with the browser. On Linux the handlers double-fork out of Chrome's process group and reparent to pid 1, so a container running without an init collects them as `<defunct>` entries — which is where this actually costs something.


route:
Closing this — the replacement flag is the right switch on paper, but it cannot be a default: it crash-loops Chrome's network service and takes the whole suite down with it.

## What the PR got right

Both halves of the premise check out against the Chrome for Testing 152.0.7977.64 Linux binary:

```
disable-crashpad-for-testing     -> 1   # real switch
no-crashpad                      -> 0   # not a Chromium switch at all
```

And it does reduce handlers (measured on `linux/amd64`, Chrome 152): no flag → 4 handler processes, `--no-crashpad` → 4 (confirming it was a no-op), `--disable-crashpad-for-testing` → 2. So `--no-crashpad` really was dead weight since 0.18.0.

## What it breaks

The switch is named `for-testing` because it is meant for Chromium's own test harness, where the process tree is fully controlled. Passed to a normally-launched browser, child processes die at startup:

```
Crashing due to FD ownership violation:
#7 0x... ChromeMain
#9 0x... __libc_start_main
ERROR:content/browser/network_service_instance_impl.cc:618]
      Network service crashed or was terminated, restarting service.
Crashing due to FD ownership violation:      <- and again, in a loop
```

`Crashing due to FD ownership violation` is Chromium's FD-ownership enforcement in `base/files/scoped_file.cc`, which hard-crashes when a descriptor is closed or acquired by code that does not own it. `ChromeMain` at frame #7 is process startup — and `ChromeMain` is the entry point for child processes too, not just the browser.

The result is the worst possible failure shape:

- the **browser process survives**, so Ferrum gets its ws URL, CDP answers normally, and the suite runs to completion
- there is **no network service**, so every request dies before leaving the browser
- every navigation returns `net::ERR_ABORTED` / `canceled: true`, the document stays `about:blank`

## Evidence from CI

Run [33176545956](https://github.com/rubycdp/ferrum/actions/runs/33176545956): **315 of 603 examples failed, on all five Ruby versions.**

Across the 253 non-empty CDP logs in the uploaded artifacts:

- 484 × `"errorText":"net::ERR_ABORTED"`
- **0 × `Network.responseReceived`** — Chrome never received a single HTTP response

Every navigation looks like this:

```
> {"method":"Page.navigate","params":{"url":"http://127.0.0.1:38571/attachment"}...
< {"method":"Network.loadingFailed","params":{...,"errorText":"net::ERR_ABORTED","canceled":true}}
< {"id":115,"result":{...,"errorText":"net::ERR_ABORTED","isDownload":false}}
< {"id":116,"result":{"root":{...,"documentURL":"about:blank"}}}
```

Every downstream failure falls out of that: `at_css` returns `nil` (`undefined method 'page' for nil`), the AX snapshot is `["RootWebArea", "none", "none"]`, and the specs expecting `Ferrum::StatusError` report "nothing was raised" because `go_to` swallows the abort.

**Why it looked like a hang rather than a failure:** `spec_helper.rb:68` takes a full-page screenshot after every failed example on CI, and each of those screenshots then times out at `FERRUM_DEFAULT_TIMEOUT=15`. 315 failures × 15s is the entire 25-minute runtime, not a deadlock.

**The environment was not the variable.** The last green run on `main` ([32970563311](https://github.com/rubycdp/ferrum/actions/runs/32970563311), 603 examples / 0 failures) and these red runs are identical: same runner image release `ubuntu24/20260823.283`, same Chrome for Testing 152.0.7977.64, same four action SHAs, same bundler cache key. The merge commit CI tested adds only `logo.svg` on top of the flag change.

## Reproduction

Controlled A/B, native `linux/arm64`, Debian trixie + apt `chromium` 151.0.7922.173, identical flags except one:

```console
$ docker run --rm --platform linux/arm64 debian-chromium bash -lc '
    chromium --headless --no-sandbox --disable-gpu --disable-dev-shm-usage \
      --user-data-dir=/tmp/ud1 --dump-dom "data:text/html,<h1>hello</h1>"'
<html><head></head><body><h1>hello</h1></body></html>          # renders instantly

$ # ... same command + --disable-crashpad-for-testing
Crashing due to FD ownership violation:
... Network service crashed or was terminated, restarting service.
TIMEOUT_EXIT=124                                                # never renders, hangs forever
```

### Repro caveat worth recording

This does **not** reproduce under emulated `linux/amd64` Docker on Apple Silicon. Four separate emulated containers — Debian bookworm, Ubuntu 24.04, and a GitHub runner mirror image, all with the exact CI Chrome build — passed cleanly, including a full 603-example run. The same Debian Chromium 151 build shows `rendered: 1, fd-violations: 0, netsvc-crash: 0` on emulated x86_64 and crash-loops on native aarch64.

So this is not architecture-specific: it fires on native aarch64 and on GitHub's native x86_64 runners, and is hidden only by emulation. Chrome process/FD behaviour has to be chased in a **native-arch** container or the repro will lie to you.

Issue #470, opened 2024-07-03
Ferrum 0.15.0 causes crash in CI and test suite collapse


pond:
**Describe the bug**

I'm not sure how to supply a useful bug report here given the bizarre behaviour but - we have a large RSpec test suite including lots of headless Chrome tests that run on AWS CI (CodeBuild / CodePipeline) triggered off GitHub commits. Recently, we updated our bundle which took Cuprite from 0.15.0 to 0.15.1; this in turn requires Ferrum 0.15.0. Our test suite started failing spectacularly, but intermittently (maybe a 70% failure rate, at arbitrary points in the suite, apparently regardless of seed). I'll explain more elsewhere in the template since this section is meant to be "brief" - we see this:

```
#<Thread:0x00007f4e63f053a8 /usr/local/bundle/gems/ferrum-0.15/lib/ferrum/utils/thread.rb:9 run> terminated with exception (report_on_exception is true):
/usr/local/bundle/gems/ferrum-0.15/lib/ferrum/client.rb:90:in `send_message': Timed out waiting for response. It's possible that this happened because something took a very long time (for example a page load was slow). If so, setting the :timeout option to a higher value might help. (Ferrum::TimeoutError)
    from /usr/local/bundle/gems/ferrum-0.15/lib/ferrum/client.rb:22:in `command'
```

Part of Ferrum appears to then have crashed, or the Chrome instance has, because *all* subsequent tests fail with:

```
Ferrum::NoSuchTargetError:
  Ferrum::NoSuchTargetError
```

**To Reproduce**

This is the problem; it replicates easily in AWS CI but we can't reproduce it locally. It seems that a `TimeoutError` comes from `client.rb` line 90, exactly as it does if I were to, say, deliberately set the `timeout` option to something very low. When I do this on local machine or in CI in an attempt to provoke replication, I just see things failing "as expected" with:

```
Ferrum::TimeoutError:
  Timed out waiting for response. It's possible that this happened because something took a very long time (for example a page load was slow). If so, setting the :timeout option to a higher value might help.
```

In case it is important - we do then notice that after a few hundred failures like this, suddenly the message changes and subsequent tests say:

```
 Errno::ENOMEM:
   Cannot allocate memory - /usr/bin/chromium
```

**I do not know if this is important or just an unrelated minor bug** arising from setting timeout so low and Ferrum perhaps not closing down old Chromium instances if its comms time out too soon, since it **occurs when we were trying but failing to replicate the nasty crash error**.

[...]

PR #633, opened 2026-09-02
change(options): stop tying :protocol_timeout to :timeout

route:
Refs #470.

## The problem

```ruby
DEFAULT_PROTOCOL_TIMEOUT = ENV.fetch("FERRUM_PROTOCOL_TIMEOUT", DEFAULT_TIMEOUT).to_i
```

`:timeout` is a user-facing knob: how long to wait for a navigation, an evaluation, a DOM query. People set it low on purpose. `:protocol_timeout` is a different thing, the budget for internal CDP bookkeeping, `Target.createTarget`, `Target.attachToTarget`, `Page.enable`, calls that resolve in milliseconds on a healthy machine.

Tying the second to the first means `Ferrum::Browser.new(timeout: 1)` also gives target creation one second, and 5 by default. On a loaded CI box that is not enough, and what comes out is a `Ferrum::TimeoutError` while creating a page, which is precisely the shape of #470: `Page.enable` never answering on a busy runner, then the whole run unravelling.

## The change

`:protocol_timeout` defaults to 30 seconds, independent of `:timeout`, still settable per browser and through `FERRUM_PROTOCOL_TIMEOUT`. Page-level waits are unaffected, `Page#command` passes `:timeout` as before.

Generous is the right default here: running out of this budget means the machine is struggling, not that the browser is gone. A browser that has actually died is detected by the socket closing, not by this timer.

## Verification

- New example in `spec/browser_spec.rb`: `timeout: 1` no longer drags `protocol_timeout` down with it, and the setter still works. Fails before the change, `protocol_timeout` is 1 there.
- Full suite: 616 examples, 0 failures.
- `:protocol_timeout` is now documented in `docs/2-customization.md` next to `:timeout`, it was missing.

Best merged after #630, which is what makes a dead browser raise immediately rather than after this timeout. On its own this change makes that detection slower, 30s instead of 5s, until #630 lands.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,9 @@
   mobile device's viewport and touch support together. [#94]
 
 ### Changed
+- `:protocol_timeout` defaults to 30 seconds instead of following `:timeout`. Internal CDP calls resolve in
+  milliseconds, so the old default made `timeout: 1` shorten browser bookkeeping too and turned a loaded machine
+  into `Ferrum::TimeoutError` during page creation [#470]
 - `Ferrum::Page::Stream#stream` now closes the CDP stream handle (`IO.close`) once it's been fully read, so streams
   opened for `Ferrum::Browser#pdf` and `Ferrum::Page::Tracing#record` no longer keep their backing storage alive in
   the browser; a failed `IO.close` is raised to the caller.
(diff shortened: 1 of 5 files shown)


route:
Closing this. 30s is far too much for a protocol call: `:protocol_timeout` only bounds browser-session bookkeeping (`Target.createTarget`, `Target.attachToTarget`), page commands go through `Page#command` with `:timeout`. Those responses are quick, what's slow is the page load, and that's a different budget. The unanswered `Page.enable` in #470 wasn't a slow response either, it never reached the browser, which #602 fixed.

